PR #1, opened 2018-06-10
Rewrite/restyle

shadycake:
Restyling and rewrite of the current bot functionality.


thebeanogamer:
Very impressed with what has been produced here. @shadycake is to be congratulated.

PR #6, opened 2018-06-24
Add Docker support

fractalpenguin:


bot/cogs/fun.py
```diff
@@ -98,7 +98,7 @@ def __init__(self, bot: Bot):
             await ctx.send(f"Unknown emojis: {emoji_string}")
 
     @command()
-    async def xkcd(self, ctx: Context, number=None):
+    async def xkcd(self, ctx: Context, number: str = None):
```
shadycake: Why is a number a string? Also kwargs shouldn't have spaces around the `=` if i recall correctly.

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3
+
```
shadycake: Add these lines to use tini
```
RUN apk add --update tini

ENTRYPOINT ["/sbin/tini", "--"]
```
Using this makes everything work more gracefully when shutting down.

thebeanogamer:
@shadycake If the requested changes were made, would this be good to merge?

thebeanogamer:
Also, as the CI files are only on the master branch, I can't run CI checks on this.

Issue #15, opened 2018-06-28
Using the help command on Unbelievaboat leads to it being muted by this bot


interiorchronos:
The bot seems to be tripping on the link that unb. is posting when the help command is called. It has to be manually unmuted by a staff member each time this happens. You can also mute the bot using it's own method, which may be worth looking into

interiorchronos:
More testing shows that it is caused by the discord link that the bot posts - is it possible to exclude all bots from being muted by the CD bot?

fractalpenguin:
Seeing as a PR solving this issue was merged, I assume it is safe to close this issue.

PR #16, opened 2018-06-29
Prevent other bots from being auto-muted

AnonGuy:
This is a minor edit that fixes issue #15.


thebeanogamer:
Does it actually work?

PR #25, opened 2018-07-01
Santized the output of :react

shadycake:


fractalpenguin:
@AnonGuy, that seems to function only for pre-existing messages. It might be better to bar the string entirely.

shadycake:
@AnonGuy I tried using that, but it wouldn't work. Also, users can mention other users anyways so it shouldn't matter surely?

PR #34, opened 2018-08-07
Add Kali reaction

picapi:
Alike the dabato reaction , react with a police car every time a user mentions Kali, because it's "illegal" apparently 👀👀

### DIFF
--- a/bot/cogs/fun.py
+++ b/bot/cogs/fun.py
@@ -33,7 +33,7 @@ class Fun:
         # React if message contains dabato.
         if "dabato" in message.content:
             await message.add_reaction("🤔")
-            
+
         # React if message contains kali.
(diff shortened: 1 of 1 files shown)


teamshortcut:
Yes except can it be the siren emoji instead (or both)

thebeanogamer:
You need to make this pass checks before I will even consider merging it

bot/cogs/fun.py
```diff
@@ -34,6 +34,10 @@ def __init__(self, bot: Bot):
         if "dabato" in message.content:
             await message.add_reaction("🤔")
 
+        # React if message contains kali.
```
thebeanogamer: Kali is a proper noun, and as such should be capitalised in this comment.

PR #35, opened 2018-08-27
Add "text to speech bubble" feature w/ Agent J and Jibhat pictures.

fractalpenguin:


bot/constants.py
```diff
@@ -1,7 +1,10 @@
+"""
+A list of constants.
+"""
 # Fun constants
 
 EVERYONE_REACTIONS = ["🙁", "🇼", "🇭", "🇴", "🔵", "🇵", "🇮", "🇳", "🇬", "🇪", "🇩"]
-QUOTE_CHANNEL_ID = 463657120441696256
```
thebeanogamer: This is the wrong id

requirements.txt
```diff
@@ -1 +1,2 @@
-https://github.com/Rapptz/discord.py/archive/rewrite.zip#egg=discord.py[voice]
\ No newline at end of file
+git+https://github.com/Rapptz/discord.py.git@rewrite#egg=discord.py[voice]
```
thebeanogamer: Why are you changing this?

bot/cogs/fun.py
```diff
@@ -99,11 +101,12 @@ def __init__(self, bot: Bot):
         if unknown_emojis:
             emoji_string = ", ".join(unknown_emojis)
             # Inserts an invisible character to render mass mentions ineffective.
-            emoji_string = emoji_string.replace("@here", "@\xadhere").replace("@everyone", "@\xadeveryone")
+            emoji_string = emoji_string.replace(
+                "@here", "@\xadhere").replace("@everyone", "@\xadeveryone")
```
thebeanogamer: Why are you changing this?

thebeanogamer:
It may also be worth adding a note to the readme.md to explain the new commands.

thebeanogamer:
Codacy seems to be hanging. If Travis passess, but Codacy is stuck, I'm fine with merging.

requirements.txt
fractalpenguin: It's the preferred way of referring to a git repository in the requirements.txt file:
https://pip.readthedocs.io/en/stable/reference/pip_install/#git

bot/cogs/fun.py
fractalpenguin: That was due to the line being over 80 characters, which goes against PEP8:
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

PR #36, opened 2018-08-28
Fixes for the quotes function

fractalpenguin:
### Bugfixes

- Fixed issue preventing quotes from a specific user from being selected when the `:quote` function was called with an username, or a mention.

### Improvements

-  Replaced constant QUOTE_CHANNEL_ID with a `:select_quote_channel` function, eliminating the need for a global variable. Note that this function is currently available to all users, so some sort of role constraint would be advised.
- Replaced the RNG from `random.choice()` with the CSRNG from `secrets.choice()`: The improved randomness of the CSRNG results in a lower likelihood of a repeat quote.


bot/cogs/fun.py
```diff
@@ -175,6 +180,14 @@ def __init__(self, bot: Bot):
         embed_quotation = quotation.embeds[0]
         await ctx.send(embed=embed_quotation)
 
+    @command()
```
thebeanogamer: This needs premissions checking otherwise it shall immediately be broken.

bot/cogs/fun.py
```diff
@@ -1,15 +1,16 @@
 """
 Set of bot commands designed for general leisure.
 """
-from random import choice, randint
+from random import randint
+from secrets import choice
```
thebeanogamer: Do we really need two random number functions?

Our random quote picker really doesn't need to use the crypto number generator.

thebeanogamer:
Also, it needs to pass Travis checks to be merged.

thebeanogamer:
It may also be worth adding a note to the readme.md to explain the new commands.

bot/cogs/fun.py
fractalpenguin: It has been added. See the line below.

thebeanogamer: I can't see a permissions check?

fractalpenguin: @has_any_role(*ADMIN_ROLES)

PR #41, opened 2018-08-29
Changes to Quote Functionallity - Possible fix to user-specific mentions.

picapi:
Changed the Quotes Channel ID to a constant, which is loaded the firs…t time the quotes command is used.

Attempted to optimise the quotes function, including adding functionality for quotations without mentions
to be functional, and avoiding errors due to the lack of embeds etc.


bot/cogs/fun.py
```diff
@@ -171,23 +171,32 @@ def __init__(self, bot: Bot):
...
         quotation_channel = self.quote_channel
         if member is not None:
-            all_quotations = await quotation_channel.history(limit=None).flatten()
             quotations = []
-            for quotation in all_quotations:
-                embed = quotation.embeds[0]
-                author_name = embed.author.name
-                author = ctx.message.guild.get_member_named(author_name)
-                if author == member:
+            async for quotation in quotation_channel.history(limit=None):
```
fractalpenguin: Should result in a more responsive bot, but it needs to be tested first.

bot/cogs/fun.py
```diff
@@ -171,23 +171,32 @@ def __init__(self, bot: Bot):
...
+                    if author == member:
+                        quotations.append(quotation)
+                elif member in quotation.mentions:
                     quotations.append(quotation)
         else:
             quotations = await quotation_channel.history(limit=None).flatten()
-        quotation = choice(quotations)
-        embed_quotation = quotation.embeds[0]
-        await ctx.send(embed=embed_quotation)
+        if len(quotations) > 0:
```
fractalpenguin: Good: this removes the chance that the bot throws an error.

bot/cogs/fun.py
```diff
@@ -14,7 +14,7 @@
 from wand.image import Image
 
 
-from bot.constants import ADMIN_ROLES, EVERYONE_REACTIONS
+from bot.constants import ADMIN_ROLES, EVERYONE_REACTIONS, QUOTES_CHANNEL_ID
```
fractalpenguin: Although the bot is just designed for a single server, having a default constant set feels wrong, somewhat.

fractalpenguin:
Might want to scrape all the quotes upon bot startup into memory, and then have an on message function listening for quotes, so that the message history does not need to be fetched every time.

bot/cogs/fun.py
picapi: There's a few different ways it can be addressed:

- Using the command in order to set the bot channel, which would need to be done every time the bot reboots, but allows more flexibility while the bot is running.
- Setting an ID in the constants file, meaning the bot needs no further configuration on boot, but may be less flexible, requiring a restart in order to change the ID.
- Setting a default, constant ID while maintaining the command that could be used to change this ID. This keeps flexibility, but also allows the bot to run without a hitch if the channel is not set.

The production build currently uses the first, whereas this set of changes incorporates the third of these options. There could well be other options worth exploring though.

fractalpenguin: We can't really have persistence with Docker containers, which is the issue.

AnonGuy: @fractalpenguin I disagree - the `QUOTES_CHANNEL_ID` seems like a good constant to have.
Storing the constant as an environment variable would still require a reboot of the bot, no matter what approach we take.
The only approach that would work with this mindset is calling `os.environ.get` or `os.environ.__getitem__` on each usage of the quotes command. This adds unnecessary overhead, and is a waste of resources compared to simply using a constant defined in `constants.py`. For our use-case, it's safe to say that we'll never change the quotes channel. On the off chance that we do, it's a fairly easy fix to make.

I do agree that these constants should be configurable, so what I could suggest is using a combination of both approaches:
```python
QUOTES_CHANNEL_ID = os.environ.get('QUOTES_CHANNEL_ID', '463657120441696256')
```
This would allow us to optionally override the constant, or fallback to a default value if the appropriate environment variable isn't found. We could also have an optional YAML configuration file, which would override constants defined here.

bot/cogs/fun.py
AnonGuy: This would cause the bot to iterate every message sent in the quotes channel if the user isn't found. Have you tested the performance implications of this?

bot/cogs/fun.py
```diff
@@ -171,23 +171,32 @@ def __init__(self, bot: Bot):
...
         if member is not None:
-            all_quotations = await quotation_channel.history(limit=None).flatten()
             quotations = []
-            for quotation in all_quotations:
-                embed = quotation.embeds[0]
-                author_name = embed.author.name
-                author = ctx.message.guild.get_member_named(author_name)
-                if author == member:
+            async for quotation in quotation_channel.history(limit=None):
+                if len(quotation.embeds) > 0:
```
AnonGuy: This could be written as:
```python
if quotation.embeds:
```

bot/cogs/fun.py
fractalpenguin: An environment variable can be specified using a .env file, which can be untracked from version control, making the bot server-agnostic.

I do, however, like the environment variable, with a predefined fallback.

picapi:
I'm planning on re-writing this at some point, to fit better with the evolving codebase around it, but won't have much time to do so in the near future.

PR #42, opened 2018-08-30
Added Docker tests to Travis

fractalpenguin:
Tests to see if a docker image can be built.


.travis.yml
```diff
@@ -6,11 +6,16 @@ branches:
   only:
     - "master"
 
+services:
+  - docker
+
 install:
-  - pip install -r requirements.txt
+  - docker build -t cd-bot .
+  - 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then docker run -d -e BOT_TOKEN --name cd-bot cd-bot; fi'
```
AnonGuy: Have you tested the use of environment variables this way? I'm not very well versed with Travis, but surely `BOT_TOKEN` should be `$BOT_TOKEN` or `"${BOT_TOKEN}"` or something of the sort.

fractalpenguin: This is directly taken from the Travis website : https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions

AnonGuy: The Travis page you linked to refers to the bash conditional. I'm not referring to the conditional, I'm referring to how you're incorrectly trying to reference the `BOT_TOKEN` environment variable. Environment variables are usually referenced with a dollar sign before the variable name, the same way the Travis example shows you. `BOT_TOKEN` should be changed to `$BOT_TOKEN`.

Here's an example of how to use environment variables this way:
image
The first example is what you're doing currently, and the second is what it should be.

AnonGuy: Ah, just read up on the documentation for the `docker run` command, looks like I misunderstood  :eyes:

thebeanogamer:
Just to check before I merge this @fractalpenguin, you are launching an instance of the bot if it's not a PR to check that it works. How long will this instance live for? Will Travis just kill the instance once it has reached the end of the file, and if so, surely we should move the docker checks to after flake8 to prevent the container running for longer than needed?

fractalpenguin:
@thebeanogamer: Yup, the instance is killed as soon as it can be started.

PR #45, opened 2018-09-09
Update admin.py with purge_from command.

SolitudalDeveloper:
Added purge_from command. Purges messages after a certain ID

### DIFF
--- a/bot/cogs/admin.py
+++ b/bot/cogs/admin.py
@@ -44,6 +44,18 @@ class Admin:
 
         await self.unmute_member(member, reason=reason)
         await ctx.send(f"{ctx.author.mention} | {member.mention} has been unmuted.")
+    
+    @command()
+    @has_any_role(*ADMIN_ROLES)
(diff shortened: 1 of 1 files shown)


thebeanogamer:
This is way too dangerous to ever be somthing that I'd allow near the bot

thebeanogamer:
I also have no idea who @SolitudalDeveloper is

SolitudalDeveloper:
@thebeanogamer fair enough, I just created it in response to an issue

Ge0rg3:
@thebeanogamer doesn't Dyno already do this?

SolitudalDeveloper:
> 
> 
> @thebeanogamer doesn't Dyno already do this?

Not specifically. It has purge commands, but not one that takes a message ID, and purges all messages since that message.

AnonGuy:
It's been almost a month, and none of the build checks have been solved. I think it's safe to close this PR and open an issue for the suggested command.

PR #48, opened 2018-10-02
Add new "FBI OPEN UP" Reaction; formatted others

bullseye2030:
Added the FBI OPEN UP reaction to the bot, with multiple trigger words, had all the reactions inline, and formatted the other multiple-emoji reacts formatted the same.


bullseye2030:
@thebeanogamer suck it and see 👇 😎 👇

AnonGuy:
Waiting on @thebeanogamer to merge this one

bot/cogs/fun.py
```diff
@@ -39,6 +39,14 @@ def __init__(self, bot: Bot):
         if "dabato" in message.content:
             await message.add_reaction("🤔")
 
+        # React FBI OPEN UP if message contains trigger words.
+        triggers_list = ["child", "fbi", "loli", "hentai", "illegal", "maltego"]
+        react_list = ["🇫", "🇧", "🇮", "⚫", "🇴", "🇵", "🇪", "🇳", "⚪", "🇺", "🅿️"]
```
AnonGuy: Might be able to use the following syntax soon, if this PR is still open after #49 gets merged:
```python
react_list = self.bot.emojify('fbi open up')
```

thebeanogamer:
Please actually test code before submitting

thebeanogamer:
I will test this later, but looking at it, all you've done is move the broken emoji to a different place, not replaced it with one that works.

bot/cogs/fun.py
```diff
@@ -39,6 +39,14 @@ def __init__(self, bot: Bot):
         if "dabato" in message.content:
             await message.add_reaction("🤔")
 
+        # React FBI OPEN UP if message contains trigger words.
+        triggers_list = ["child", "fbi", "loli", "hentai", "illegal", "maltego"]
+        react_list = ["🇫", "🇧", "🇮", "⚫", "🇴", "🅿️", "🇪", "🇳", "⚪", "🇺", "🇵"]
+        for _ in range(0, len(triggers_list) - 1):
```
shadycake: I'd suggest not using `_`, since it's conventionally used as a throwaway variable. Use something like `trigger`.

bullseye2030: I will do

bot/cogs/fun.py
```diff
@@ -39,6 +39,14 @@ def __init__(self, bot: Bot):
         if "dabato" in message.content:
             await message.add_reaction("🤔")
 
+        # React FBI OPEN UP if message contains trigger words.
+        triggers = ["child", "fbi", "loli", "hentai", "illegal", "maltego"]
+        reactions = ["🇫", "🇧", "🇮", "⚫", "🇴", "🅿️", "🇪", "🇳", "⚪", "🇺", "🇵"]
+        for trigger in range(0, len(triggers) - 1):
+            if triggers[trigger] in message.content:
+                for reaction in range(0, len(reactions) - 1):
```
shadycake: Actually, not sure how I missed this but why are you iterating over a range and then using that as an index? Just iterate over the lists.

fractalpenguin: In addition, instead of having an outer loop to iterate over trigger words, it can be replaced with `any()`:
`if any(trigger in message.content.lower() for trigger in triggers)`

`any()` has the advantage of immediately exiting as soon as one of the triggers is found, rather than continuing to find more triggers.

AnonGuy:
image

PR #60, opened 2018-11-07
No u

ed588:
When the bot is mentioned, it will react to that message with the `NoU` emote

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -182,3 +182,5 @@ fabric.properties
 # Editor-based Rest Client
 .idea/httpRequests
 
+# VSCode
+.vscode/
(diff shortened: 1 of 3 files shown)


AnonGuy:
There are a few reasons why we don't need this PR:
* It creates a new, unnecessary bot cog for an arbitrary feature.
* It uses **`discord.utils.find`**, when it should be using **[`discord.utils.get`](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.utils.get)**.
* It searches all emojis in the server list on every message, when it should be defining the emoji elsewhere.
* It implements a feature that wasn't discussed in an issue, or with the community, beforehand.

That being said, I am closing this PR. Feel free to create a new pull request if you can address those issues.

Issue #64, opened 2018-11-09
Make The Staff Hate Us Less


thebeanogamer:
**Is your feature request related to a problem? Please describe.**
The staff keep getting annoyed with us for tagging them.

**Describe the solution you'd like**
When the Staff role is tagged, the bot should repond with `Have you tried emailing support@joincyberdiscovery.com? Please make sure to do that before unnecessarily pinging the staff!`. It could also respond to if the staff are individually tagged, but this would require a cooldown to prevent spam.

**Describe alternatives you've considered**
Screaming at people to stop doing it.

**Additional context**
To help with development, here are some IDs:
`Official Cyber Discovery Staff: 510507184396959785`
`@ampersand#3620: 453547861888598018`
`@Flex#0394: 474332857783943169`
`@Hlant#2066: 499567665774329859`
`@jibhat#1294: 453547861888598018`

thebeanogamer:
This is something the staff have specifically requested: 
![image](https://user-images.githubusercontent.com/3729048/48278885-bf1d1d00-e446-11e8-8a89-631d3d47f567.png)

AnonGuy:
After some discussion, we've come up with some possible solutions:

- [ ] **Create a fake, pingable staff role to lure help vampires**
Follow the fake ping with a message like "Have you tried emailing support?"

- [ ] **Discourage users from pinging staff individually**
Maybe this could be done by sending a similar message, or muting users who abuse staff pings?

thebeanogamer:
With Game coming up, this needs to be urgently resolved. Personally, my prefered solution is #79, although I guess that #65 would probably also work.

If someone is able to test #79 and confirm that it works, then I think it's good to merge.

AnonGuy:
I'll be making a PR for this tonight, since #79 has been inactive for a while.

teamshortcut:
👀

PR #65, opened 2018-11-09
Make the Staff Stop Hating Us

kaovd:
I actually have no idea if this will work someone take over

### DIFF
--- a/bot/cogs/fun.py
+++ b/bot/cogs/fun.py
@@ -89,6 +89,11 @@ class Fun:
         if "revive" in message.content.lower():
             await emojify(message, "nou")
 
+        # Stop the staff from hating us.
+        if "@Official Cyber Discovery Staff" in message.content:
+            await message("Have you tried emailing support@joincyberdiscovery.com? Please make sure to do that before unnecessarily pinging the staff!")
+
+
     @command()
     async def lmgtfy(self, ctx: Context, *args: str):
         """


thebeanogamer:
When implemented this will fix #65

Ge0rg3:
What's the point in this though? This warns the user about pinging the staff, but the ping still goes through as far as I can see. Wouldn't it be better if we got the user to confirm their intent, ie with a reaction/reply?

joker314:
Can bots edit messages?

If yes, I suppose we could have a reaction confirmation thing. But if not, how can we get a user to confirm their intent? Don't we have to delete their message, and so when they decide to keep it, the authorship information will be stripped. (And even if the bot reposts the message and notes who authored it, the user now lacks the ability to delete the message etc., especially when (if) the bot goes offline)

Bentechy66:
> This seems like a good idea @Ge0rg3, I'm just wondering if bots can edit messages?
> 
> If yes, I suppose we could have a reaction confirmation thing. But if not, how can we get a user to confirm their intent? Don't we have to delete their message, and so when they decide to keep it, the authorship information will be stripped. (And even if the bot reposts the message and notes who authored it, the user now lacks the ability to delete the message etc., especially when (if) the bot goes offline)

Bots cannot edit messages; the Discord API provides no way of doing so. Even if it did, if one pings and then edits one's message to remove the ping, the person who was pinged will still get a notification.

One solution would be to have a dummy role which is called "Cyber Discovery Staff", which none of the staff actually have, and make the real staff role non-pingable. Then, when someone pings the dummy role, the bot asks them to confirm their intent and once they do so, temporarily makes the real role pingable and pings it, then disables the pingable feature of the real role. It could also ping staff members individually.

AnonGuy:
I agree with @Bentechy66. Is there anything that you need help with before you attempt this, @kaovd?

joker314:
Of course, there is the issue that this method can only handle one person at a time. If one person confirms, but the other denies, then they will both mention the relevant administrator.

thebeanogamer:
Closed as we've decided to use a system in the style of #64 instead.

PR #67, opened 2018-11-20
Assess 0 is not a challenge!

AvrooVulcan:
Just made :assess 0 not work - that's all!

~~no assess 0 for you all!~~

### DIFF
--- a/bot/cogs/cyber.py
+++ b/bot/cogs/cyber.py
@@ -81,7 +81,7 @@ class Cyber:
         with open("assess.json") as f:
             assess_docs = load(f)
 
-        if challenge_num not in range(len(assess_docs) + 1) and challenge_num != 0:
+        if challenge_num not in range(1, len(assess_docs) + 1):
             await ctx.send("Invalid challenge number!")
 
         else:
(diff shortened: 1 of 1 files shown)


bot/cogs/cyber.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, bot: Bot):
         with open("assess.json") as f:
             assess_docs = load(f)
 
-        if challenge_num not in range(len(assess_docs) + 1):
+        if challenge_num not in range(len(assess_docs) + 1) and challenge_num != 0:
```
AnonGuy: I'd prefer it if you included the lower limit in the `range` function:
```suggestion
        if challenge_num not in range(1, len(assess_docs) + 1):
```

AnonGuy: On second thoughts, why are we even calling the `range` function in the first place?
```suggestion
        0 < challenge_num <= len(access_docs):
```

bot/cogs/cyber.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, bot: Bot):
         with open("assess.json") as f:
             assess_docs = load(f)
 
-        if challenge_num not in range(len(assess_docs) + 1):
+        if 0 < challenge_num <= len(access_docs):
```
AnonGuy:
```suggestion
        if 0 < challenge_num <= len(assess_docs):
```

PR #75, opened 2018-11-29
 Add URL attribute to assess embed.

aquariuscd:


thebeanogamer:
Isn't this just undoing #75? Why was that merged if it's going to be instantly reverted?

Do I need to increase the number of approving reviews required?

PR #76, opened 2018-11-30
Optimise quotes system by implementing a cache in memory

aquariuscd:
Currently, the quotes command searches the entire history of the quotes channel every time someone runs :quotes, this PR implements an in memory cache so the command is much faster. This was originally going to be a SQLite database but Aperture and I decided it would be too much effort for too little performance payoff compared to this.

### DIFF
--- a/bot/__main__.py
+++ b/bot/__main__.py
@@ -1,5 +1,6 @@
 """Main script to define bot methods, and start the bot."""
 
+from collections import defaultdict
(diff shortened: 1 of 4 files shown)


AnonGuy:
I will be testing this locally before merging.

AnonGuy:
I've tested this on my own server locally and it works perfectly.

PR #82, opened 2018-12-09
fixing username changes breaking some quotes

justanotherfailure:
fixing username changes breaking quotes

### DIFF
--- a/bot/cogs/fun.py
+++ b/bot/cogs/fun.py
@@ -62,8 +62,12 @@ class Fun:
     async def on_message(self, message: Message):
 
         if message.channel.id == QUOTES_CHANNEL_ID and message.author.id == QUOTES_BOT_ID:
-            author = message.embeds[0].title
-            self.bot.quotes[author].append(message.id)
+            author = message.embeds[0].author.icon_url
+            if 'avatars' not in author:
+                print('A quote was unable to be cached due to author having a default avatar.')
+            else:
+                author = int(author.split('avatars')[1][1:19])
+                self.bot.quotes[author].append(message.id)
 
         """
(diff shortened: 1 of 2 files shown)


joker314:
Note to reviewers: it appears the line endings have changed. This may not be desirable, and might need to be fixed, but in the mean time, select "Ignore whitespace" in the diff settings to get a more manageable diff.

justanotherfailure:
> 
> 
> Note to reviewers: it appears the line endings have changed. This may not be desirable, and might need to be fixed, but in the mean time, select "Ignore whitespace" in the diff settings to get a more manageable diff.

This is due to me simply uploading and replacing the files on my local repo, had i used normal git commands this probably wouldn't have happened, apologies.

bot/cogs/fun.py
```diff
@@ -1,257 +1,257 @@
...
+from itertools import cycle
+from random import choice, randint
+from string import ascii_lowercase
+from typing import AsyncGenerator
+from urllib.parse import urlencode
+
+from aiohttp import ClientSession
+from discord import Embed, File, Message
+from discord.ext.commands import (
+    Bot, Context, TextChannelConverter, command, has_any_role, UserConverter
```
joker314: We put the uppercase things first, then the lowercase things, and each is sorted alphabetically. In this case,
```suggestion
    Bot, Context, TextChannelConverter, UserConverter, command, has_any_role
```

justanotherfailure:
It seems that I hadn't remembered to cache new quotes while the bot was running, I've added that now with my last 2 commits.

justanotherfailure:
imma close this, and make a proper one without broken line endings

PR #92, opened 2019-01-02
Give Cyberstart Essentials Some Love

thebeanogamer:
In the style of the `:game` command, this PR adds a `:essentials` command to see the start date of essentials and how long is left until then.


joker314:
I really like this! Would it be a good idea to make a function instead of repeating the same code twice; as that would make it more maintainable?

PR #95, opened 2019-01-03
Fixing automatic response regex bug

Ge0rg3:
Currently, a user can have all of the above regex in one message, and thus the bot would spam replies. This should prevent that, by only responding to the first match.


JosephBywater:
example: 
assess start assess end game start game end essentials start essentials end

teamshortcut:
Looks good to me. Cooldown would also help spam, but not the purpose of this PR.

Ge0rg3:
Thanks guysss

PR #100, opened 2019-01-08
Prevent bot responding to a command (again)

AnonGuy:
Same diff as #98


bot/cogs/cyber.py
```diff
@@ -271,6 +271,9 @@ def suffix_from_number(num):
         await ctx.send(f"That's in {month_and_day_countdown}!")
 
     async def on_message(self, message: Message):
+        ctx = await self.bot.get_context(message)
```
thebeanogamer: Please add a note to explain what this does for the benefit of future authors

PR #101, opened 2019-01-12
Refactor the ":game" and ":essentials" commands so that they both point to the same code

joker314:
Previously, the `game` and `essentials` commands had a copy of the same logic in their implementations. This means that making changes to that logic requires changing two things. This reduces developer efficiency and can lead to mistakes.

This pull request abstracts away this logic into the `command` function. It takes the following parameters
- The string date -- used both in the output to the user, and also parsed into a `datetime.date` to become the target for the countdown
- The name of the stage ("CyberStart Essentials" or "CyberStart Game")
- The context

This pull request was tested in the following way:
- Run the bot locally on a private server
- Observe that `:game` produces the same output as before
- Observe that `:essentials` produces the same output as before

This pull request only refactors code, and does not change behaviour for the end-user.

### DIFF
--- a/bot/cogs/cyber.py
+++ b/bot/cogs/cyber.py
@@ -212,10 +212,18 @@ class Cyber:
 
     @command()
     async def game(self, ctx: Context):
+        """
+        Gets the date of, and days and months until, CyberStart Game
+        """
+
         await self.countdown('15th January 2019', 'CyberStart Game', ctx)
 
     @command()
     async def essentials(self, ctx: Context):
+        """
+        Gets the date of, and days and months until, CyberStart Essentials
+        """
+
         await self.countdown('5th March 2019', 'CyberStart Essentials', ctx)
 
     async def countdown(self, countdown_target_str: str, stage_name: str, ctx: Context):


AnonGuy:
I've tested the bot locally, and can confirm that the functionality works fine.
image

bot/cogs/cyber.py
```diff
@@ -211,63 +212,40 @@ def __init__(self, bot: Bot):
...
-        if today > game_start_date:
-            await ctx.send("Cyberstart Game has begun! Use :level base level to get info"
-                           "on specific challenges once we update the bot")
-            return
-        await ctx.send("Cyberstart Game begins on the 15th January 2019.")
-        await ctx.send(f"That's in {month_and_day_countdown}!")
+        await self.countdown('15th January 2019', 'CyberStart Game', ctx)
 
     @command()
     async def essentials(self, ctx: Context):
```
AnonGuy: Add a docstring explaining the command. The `:help` function uses this docstring to generate a description, see my screenshot above.

bot/cogs/cyber.py
```diff
@@ -211,63 +212,40 @@ def __init__(self, bot: Bot):
 
     @command()
     async def game(self, ctx: Context):
```
AnonGuy: Add a docstring explaining the command. The `:help` function uses this docstring to generate a description, see my screenshot above.

joker314:
I've added docstrings for both of those commands, and also tested that the `:help` command works with these docstrings. Thanks for suggesting this @AnonGuy 

In addition, I confirmed that the `:level` command still uses `:game` as a fallback correctly.

PR #103, opened 2019-01-13
Show a warning when requesting an assess challenge >8

joker314:
The bot allows users to read the challenges of the CyberStart Assess stage using the `:assess` command. However, other users are not allowed to provide hints for challenges 9 through 14, so as to keep the competition fair.

Despite the #assess channel clearly asking users to read the pinned messages, as an extra barrier against people asking for hints, this pull request introduces a warning message. It looks like this:

> **:warning: Remember, we can't give hints after challenge 8**

It appears directly before the challenge description in the bot's response body, if a challenge number greater than 8 is requested.

The hints limit is placed in a variable so it can be easily updated in the future.

Here is a screenshot of the bot running.

Responses to requests for challenge 9 include the warning

Responses to requests for challenge 8 do not include the warning

### DIFF
--- a/bot/cogs/cyber.py
+++ b/bot/cogs/cyber.py
@@ -119,6 +119,9 @@ class Cyber:
         Gets information about a specific CyberStart Assess level and challenge.
         """
 
+        HINTS_LIMIT = 8  # Challenge number up to where hints are allowed (inclusive)
+        NO_HINTS_MSG = f"**:warning: Remember, we can't give hints after challenge {HINTS_LIMIT}**"
+
         # Gather Assess data from JSON file.
         with open("bot/data/assess.json") as f:
             assess_docs = load(f)
@@ -132,6 +135,10 @@ class Cyber:
             challenge_title = challenge_raw["title"]
             challenge_difficulty = challenge_raw["difficulty"]
             challenge_text = challenge_raw["description"]
+
+            if challenge_num > HINTS_LIMIT:
+                challenge_text = NO_HINTS_MSG + '\n' + challenge_text
+
             embed = Embed(
                 title=f"CyberStart Assess Challenge {challenge_num} - {challenge_title}",
                 description=challenge_text,


bot/cogs/cyber.py
```diff
@@ -119,6 +119,9 @@ def __init__(self, bot: Bot):
         Gets information about a specific CyberStart Assess level and challenge.
         """
 
+        HINTS_LIMIT = 8  # Challenge number up to where hints are allowed (inclusive)
```
AnonGuy: I understand why you'd have this variable, but I don't think this will ever be changed from 8, considering we got it from a staff member's statement:
image
So I think it's safe to just replace the `challenge_num > HINTS_LIMIT` below with `challenge_num > 8`

AnonGuy: Ditto about the conditional block below. Feel free to just replace it with a string literal.

joker314: Mmhm! I guess I was thinking more about next year

joker314:
@MatthewIsAwesome Thanks for the review! I agree with you that 'we' is not specific.

As for the variable names, they don't "have" to be capital -- variable naming is really just a convention.

But by convention, constants (variables which don't change their values) are written in capital letters. This is mainly to distinguish them from variables. The advantage of making it obvious what's a constant and what's a variable is that if you see an assignment to a constant, you can be certain you will know the value of the constant if you see it later in the computer program. Does this answer your question?

bot/cogs/cyber.py
joker314: Additionally, we still have to hard-code it in two places (once for the `if` statement, and once for the output) and that feels icky, even if we're 99% sure it won't change

AnonGuy: That's a fair point actually.

AnonGuy: This constant should probably be moved to `constants.py`, with a comment describing what it's for

PR #122, opened 2019-01-25
Implement Discord logging handler

### LINKED ISSUE #114: Implement a logging system
This could be done with Python's builtin `logging` library, `loguru` or `sentry`, but preferrably a logging system implemented from scratch (subclassed from `logging.Logger`).

It should be able to send all log messages to a channel ID specified in `constants.py`.

jos-b:
I've implemented a logging handler which can be used to send log messages from Python to Discord.

Since this requires that the asyncio event loop is running it will drop any messages which are not sent while the bot is online.

I have assigned the `logging.Logger` instance with the `DiscordHandler` to the `bot.log` attribute so you can call all relevant logging methods through that.

Examples:
```py
>>> bot.log.info("Hello world, this is info!")
>>> bot.log.error("oh NO eRRoR oh NO")
```

Embeds contain some metadata and currently look like this:
image

Closes #114.


bot/log.py
```diff
@@ -0,0 +1,58 @@
...
+}
+
+
+class DiscordHandler(logging.Handler):
+    """
+    A class implementing logging.Handler methods to send logs to a Discord channel.
+
+    >>> handler = DiscordHandler(client) # Where client is your instance of
+    ...                                  # discord.Client or commands.Bot
+    >>> logger.addHandler(handler)
```
AnonGuy: It's unlikely that we'll need to use your `DiscordHandler` outside this PR, so it's probably fine to leave out these doctests.

jos-b: Right, it wasn't intended to be "tested" (it'd break anyway, no definition of client or logger). It was more to demonstrate how it would be added as a logger (& the arguments for that). I can remove now.

jos-b: Addressed in 1c8cf8307d6a10540fb4a0c7c979cf72bc07abc0

bot/log.py
```diff
@@ -0,0 +1,54 @@
...
+        return LEVEL_COLORS.get(level_number)
+
+    def emit(self, record):
+        if not self.client.loop.is_running():
+            # The event loop is not running (discord is not connected) so
+            # do not send the message
+            return
+
+        # Fetch the logging channel
+        log_channel = self.client.get_channel(LOGGING_CHANNEL_ID)
```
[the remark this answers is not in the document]
jos-b: Can diddly-doodly.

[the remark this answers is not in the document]
jos-b: Addressed in d6129a9d32c46d9a2366b16ccc79e4c020371b91

jos-b:
The FitnessGram™ Pacer Test is a multistage aerobic capacity test that progressively gets more difficult as it continues. The 20 meter pacer test will begin in 30 seconds. Line up at the start. The running speed starts slowly, but gets faster each minute after you hear this signal. beep A single lap should be completed each time you hear this sound. ding Remember to run in a straight line, and run as long as possible. The second time you fail to complete a lap before the sound, your test is over. The test will begin on the word start. On your mark, get ready, start.

PR #123, opened 2019-01-26
Help lazy users using the :flag command

thebeanogamer:
Help lazy users using the `:flag` command by assuming that the base desired is HQ if none is specified


bot/cogs/cyber.py
```diff
@@ -132,6 +132,10 @@ def __init__(self, bot: Bot):
     @command()
     async def flag(self, ctx: Context, base: str, level_num: int, challenge_num: int = 0):
         """Generate a flag for the specified base, level and challenge."""
+        if challenge_num == 0:
+            challenge_num = level_num
+            level_num = int(base)
```
joker314: Wouldn't this sometimes throw an error (if this `base` is maliciously set to a non-`int`)? The bot automatically restarts, but could this be a DOS for a few seconds?

joker314: Now that we have error logging, could this instead be used to spam?

fractalpenguin: Not really, it's not too significant.

PR #124, opened 2019-01-26
When member joins with default avatar bot responds with link to Discord help

AlphexGaming:


bot/constants.py
```diff
@@ -63,3 +64,10 @@
     "\U0001f1ff\U0001f4a4",  # Z
     "\u26ab\U0001f535\U0001f534\u26aa"  # Whitespace alternatives
 ]
+# List of Default Discord Avatars
+
+DEFAULT_AVATARS = ["https://cdn.discordapp.com/embed/avatars/0.png",
```
joker314: All the elements in this array seem very similar. If Discord decides to change its domain to discord.com (which I think they now own), we'd need to make 4 changes here. Or, if we discover that 6.png also represents a default avatar, we would need to copy and paste an entire line. How about something like this? 

```python
DEFAULT_AVATAR_FSTRING = "https://cdn.discordapp.com/embed/avatars/%d.png"

DEFAULT_AVATARS = [DEFAULT_AVATAR_FSTRING % i for i in range(1, 6)]
```

Others may be able to advise on a prettier way to do that, the above is just illustrative. Ideally, we'd also have constants at the top which refer to the CDN domain name and the main domain name so that other constants can leech off them, but for now I think this is good enough.

AnonGuy: Could we probably have a regex pattern for this?
```python
DEFAULT_AVATAR = r'https://cdn.discordapp.com/embed/avatars/\d.png'
```
```python
if re.match(DEFAULT_AVATAR, member.avatar_url):
```

bot/cogs/fun.py
```diff
@@ -88,6 +88,15 @@ def __init__(self, bot: Bot):
         if "revive" in message.content.lower():
             await emojify(message, "nou")
 
+    async def on_member_join(self, member):
+        channel = self.bot.get_channel(JOIN_CHANNEL_ID)
+        if member.avatar_url in DEFAULT_AVATARS:
+            text = f'{member.mention} | "*grumbles* <https://support.discordapp.com/hc/en-us/articles/204156688>"'
```
AnonGuy: We could probably keep the link embed here.

thebeanogamer: Yeah, people are more likely to click the link if they know what it is

teamshortcut:
This is alienating to new users and shouldn't be a feature of the bot.

thebeanogamer:
image

PR #125, opened 2019-01-29
Ensure that the :flag command always generates unique results; and add some validation to it.

joker314:
Consider these two commands

```
:flag HQ 1 13
:flag HQ 2 12
```

Previously, these would generate the same flag. This is because the algorithm for determining a seed for the random number generator was this: `H + 1 + 13`. This is the same as `H + 2 + 12` because they both sum to `H + 14`.

The new algorithm is to consider the base-128 number with the digits `H, 1, 13` (and `H, 2, 12`). These are then converted into base 10. This guarantees uniqueness because all components will be less than 128.

*see bottom section, the actual new algorithm includes more than just the first letter
<hr>

Consider these two commands

```
:flag HQ 1 13
:flag hq 1 13
```

These would previously generate different flags, even though they are referring to the same challenge. This is because the sum `H + 1 + 13` is different to `h + 1 + 13`, as `H` and `h` have different ASCII codes.

So we create a new function which takes existing logic which we can reuse here. We call this function `get_area_from_base_abbrv`. It takes in something like `f` or `python` or `moon` or `hq` and outputs a canonical value (`Headquaters`, `Moonbase`, `Forensics`). This forces the first letter to be uppercase as a side-effect, which resolves this issue

<hr>

~~One thing we are assuming here which makes the code brittle is that all the bases have different initial letters (`H`, `M`, and `F`). Ideally, we would use every single letter of the base so that if, in future, something changes, we can still have unique flags.~~

Every single letter is now a digit of the base-128 number, so that `Headquarters` and `Hacking` bases (latter is fictitious) will result in different flags.

### DIFF
--- a/bot/cogs/cyber.py
+++ b/bot/cogs/cyber.py
@@ -94,13 +94,10 @@ class Cyber:
         elif challenge_num == 0:
             await ctx.send("Invalid challenge number!")
             return
-        # Find out which base the user is refering to.
-        for area in BASE_ALIASES.keys():
-            if base.lower() in BASE_ALIASES[area]:
-                base = area
-                break
-        else:
-            await ctx.send("Unknown base.")
+
+        base = self.get_area_from_base_abbrv(base)
+
+        if base is None:
             return
 
         # Check to see if that many levels are present
@@ -138,7 +135,7 @@ class Cyber:
             # Generates random, but unique and identical per challenge, base 64 "flag"
             #
             # First, we must generate a numeric seed (deterministically) out of:
-            # - the base (HQ/Forensics/Moon)
+            # - the area [base] (HQ/Forensics/Moon)
             # - the level [1..13]
             # - the challenge [1..13]
             #
@@ -151,12 +148,13 @@ class Cyber:
             # gain a single numeric seed which we can then pass for further computation
             # to `generatebase64`
             #
-            # To convert the base (HQ/Forensics/Moon) into a number, we take the ASCII
+            # To convert the area (HQ/Forensics/Moon) into a number, we take the ASCII
             # value of the first character present. ("H", "F", and "M" respectively).
-            #
-            # TODO: Validate that only those bases (HQ/Forensics/Moon) are allowed
 
-            seed = ord(base[0].lower()), level_num, challenge_num
+            area = self.get_area_from_base_abbrv(base)
+
+            seed = ord(area[0]), level_num, challenge_num
+
             num_base = 128  # the positional number base to use (must be > than all components of seed)
 
             aggregate_seed = sum(digit * (num_base ** digit_index) for digit_index, digit in enumerate(seed))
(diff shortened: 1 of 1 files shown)


thebeanogamer:
Whilst this is likely true, given that it is a joke command, it is not exactly the end of the world if it produces the same value sometimes. If you want to fix it, you can, but I’d consider it low priority.

fractalpenguin:
This adds a lot of unneeded complexity to what is a function with no real purpose beyond humour.

joker314:
Thanks for the feedback. I think I agree, but I don't think each feature is more than a couple of lines; and it's really the bulky comment which fills up most of the code. If you still disagree, feel free to close this Pull Request.

swedg:
This is a bit over-engineered for a joke, wouldn't you say? I get your thinking but you could've just squared the numbers before summing them or something easier if you really wanted to remove the 1 + 2 + 3 = 2 + 2 + 2 thing.

swedg:
Oh also that comment is wayyyyyyyy too long

fractalpenguin:
This function is completely useless, and is probably actively detrimental, given that people ask about it on the server, sometimes believing it to be genuine.

fractalpenguin:
If you want an unique set of characters, just treat the numbers as strings, concatenate them, then take the base64 of a hashing function e.g MD5 applied to the concatenated string.

joker314:
Hi @swedg ! It turns out the squaring the numbers just moves the problem. Sure, that specific example no longer works. But there'll be another clash. Consider this:

H<sup>2</sup> + 2<sup>2</sup> + 3<sup>2</sup> = H<sup>2</sup> + 3<sup>2</sup> + 2<sup>2</sup>

This relies on the fact that addition is _commutative_, `a + b = b + a`. In this case, we notice that `HQ L2 C3` produces the same flag as `HQ L3 C2`. This is why just squaring each term doesn't quite work. More generally `x La Cb = x Lb Ca`.

@fractalpenguin This is a good idea, as hash results are already integers which can be quickly used to seed the PRNG! You may want to file a follow-up PR as I'm closing this one.

joker314:
Despite the actual code changes here being tidy,  people seem to be fixating on the massive write-up. This is unfortunate, because comments are very important and are often underused. However, I'm not enjoying maintaining this PR and will not invest the effort into dealing with the merge conflicts of #131 .

This PR has been structured as multiple commits so if the validation wants to be kept, you have my permission to cherry-pick it and maintain the pull request further (filed separately).

joker314:
@fractalpenguin This command is being actively maintained. I agree it has ethical flaws, I encourage you to voice them in the server/with the mods; but this PR probably isn't the best place because the command itself has active moderator approval (see #131 made by a moderator)

swedg:
@joker314 Yeah my bad didn't think about that (the commutativity thing). Also it's true that a lot of people under-use comments, but that doesn't mean its not possible to over-use comments. Also this isn't fixation on the comment, we mentioned it's a complex and over-engineered answer to a simple problem, a problem that honestly doesn't really matter.

PR #127, opened 2019-02-02
Allow self assignable roles

joker314:
The current command `?add-role "Year X"` is broken. This is likely because you have to actually @mention the role -- which would lead to a lot of unwanted notifications from that year group.

Instead, we introduce the `:year` command. It is rate limited to 1 request per 5 seconds per user. This is because setting a year group is usually a fairly infrequent action (once a year, ideally); and there is a rate limit from Discord's side in place I think.

In the constants file, there's a dictionary containing all the role names ("Year 1[0123]", "Always just a bit too old"). These are the roles which are self-assignable.

The following commands work:

```
# As you'd expect
:year 10
:year 11
:year 12
:year 13

# Always just a bit too old
:year old

# Remove year-group role
:year remove
```

The command is case-insensitive, and validates its input (if invalid year group, all possible year groups listed), but throws an error if a role does not exist in the guild.

The documentation in the #readme channel will need to be updated if this is merged

<hr>

Screenshot of the bot responding to requests for year 10, 14 (which resulted in an error message), and year old -- and the 'd' in old is capitalized to show that the command is case-insensitive

### DIFF
--- a/bot/cogs/general.py
+++ b/bot/cogs/general.py
@@ -34,7 +34,9 @@ class General:
         role_name = SELF_ROLE_NAMES.get(year_group, None)
 
         if role_name is None:
-            await ctx.send(f"Not a self-assignable role. Try one of: {', '.join(SELF_ROLE_NAMES.keys())}")
+            msg = f"Not a self-assignable role. Try one of: {', '.join(SELF_ROLE_NAMES.keys())}, "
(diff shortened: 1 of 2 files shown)


joker314:
The bot seems to be very slow at dealing with such requests, but it might just be my internet connection.

joker314:
Closing in favour of recently fixed `.rank` command in another bot.

PR #131, opened 2019-02-07
Changed flags to be marked as spoilers

teamshortcut:
Just surrounded the flag with "||"

### DIFF
--- a/bot/cogs/cyber.py
+++ b/bot/cogs/cyber.py
@@ -140,7 +140,7 @@ class Cyber:
             content = "13.1 is a No Flag Zone™ 🙅⛔⚔️"
         else:
             # Generates random, but unique and identical per challenge, base 64 "flag"
-            content = "The flag is: " + await generatebase64(ord(base[0]) + level_num + challenge_num)
+            content = "The flag is: ||" + (await generatebase64(ord(base[0]) + level_num + challenge_num)) + "||"
 
         embed = Embed(
             title=(f"{base} - Level {level_num} Challenge {challenge_num}"),


bot/cogs/cyber.py
```diff
@@ -140,7 +140,7 @@ def __init__(self, bot: Bot):
             content = "13.1 is a No Flag Zone™ 🙅⛔⚔️"
         else:
             # Generates random, but unique and identical per challenge, base 64 "flag"
-            content = "The flag is: " + await generatebase64(ord(base[0]) + level_num + challenge_num)
+            content = "The flag is: ||" + (await generatebase64(ord(base[0]) + level_num + challenge_num)) + "||"
```
joker314: I don't think this extra pair of brackets helps readability, do you agree?

fractalpenguin: It doesn't exactly harm it, and it has already been merged.

joker314: > It doesn't exactly harm it

Well, that depends on what you mean by "harm".

First, let's consider the functionality of the program. For the same inputs, we can definitely expect the same outcomes. This is because parenthesis don't change the behaviour of the program. But parenthesis also take more time to parse and evaluate by the Python interpreter. This effect is negligible so we can discount it also. Hence I concede that the functionality is the same.

Let us next consider _style_. Some argue that style is unimportant. A program that runs is a good program. Yet consider that the majority of a developer's time will be _reading_ code, not writing it. So we need to make sure that the code our contributors are reading is clear (the people who contribute are volunteers, after all), and we even have certain validations in place (like `flake8`), to make sure our code is consistent. Sometimes, parenthesis are a really good thing. They help us group things together so that it is obvious how things are going to be run. This in turn makes reading code faster. Yet there are other times when parenthesis are pointless. Consider this:

```python
(24 * (60 * 60)) * 365
```
Each operand has equal weight in the expression, yet we are pointlessly grouping some together, just to create confusion. In this case, `flake8` should not be required to complain, because it doesn't know what `60` or `24` mean to humans, and hence it doesn't know which (if any) are more important than the others. This is a style decision, but for humans.

The case we're dealing with here is significantly less extreme. We have a function invocation and it's been surrounded in brackets:

```python
... (f(...)) ...
```

My criticism is that
  - no brackets are needed because there is only one term inside
  - the opening bracket isn't needed because a syntax highlighter will separate (using colour) the string literal from the function invocation
  - the closing bracket isn't needed because there's already a closing bracket to keep things isolated: the bracket of the function invocation. I find double brackets a little clunky when they aren't needed.

Effectively, this is a personal preference. And that's why I've left this open for discussion. It's just I worry that the brackets will cause a contributor to ponder why they were placed there (i.e., maybe there was a special reason that they haven't figured out yet), or they will be extra characters that distract from the content of the code.

I don't know. What are you thoughts?

> it has already been merged

A second pull request can always be filed if needed. \:) It would be a great project for a person inexperienced in version control or Python -- and I know CyberDiscovery has loads of them! We could file an issue and advertise it on the server to get more people contributing, if we wanted to.

<hr>

Whoops I rambled! My point is that this is incredibly unimportant and that I agree with you completely.

joker314: Also, we could use f-strings and that would probably be the best solution?

I'm thinking something like this:

```python
flag = await generatebase64(...)
content = f"The flag is: ||{flag}||"
```

fractalpenguin: f-strings would probably be a better solution.
The parentheses help in the sense that the await call is insulated from the rest of the program.

joker314: @fractalpenguin If you're interested in my reasoning, you could view the edit history of the removed comment. I think I agree with you. Per my deleted comment, I guess we could file a help wanted issue and reserve it for a first-time contributor to help them get started with Git and Python?

AnonGuy: Having a discussion here is pointless, since the PR is closed.

AnonGuy: Having a discussion here is pointless, since the PR is closed.

PR #135, opened 2019-02-07
add game channel regex function

maishams:
instead of having to run the :level command, it will look for 'lvl' or 'level' in a message, and then show the level thing instead. it also checks it against what channel it was sent in, so person wont have to specify what base. i tried it on a private server, and it assigned the variables well.


bot/cogs/cyber.py
```diff
@@ -78,11 +82,25 @@ def __init__(self, bot: Bot):
...
             return
 
+    async def on_message(self, message):
+        # Check the current command context
+        ctx = await self.bot.get_context(message)
+        # If message is a command, ignore regex responses.
+        if ctx.valid:
+            return
+        if self.game_level_regex.match(message.content):
+            level_num, challenge_num = int(match.group(2)), int(match.group(3))
```
joker314: So `match` isn't defined here. `self.game_level_regex.match(message.content)` indeed returns a match, but it's not being saved in a variable. Try doing

```python
match = self.game_level_regex.match(message.content)
if match:
  # do stuff with the match
```

instead?

bot/cogs/cyber.py
```diff
@@ -78,11 +82,25 @@ def __init__(self, bot: Bot):
         Gets information about a specific CyberStart Game level and challenge.
         If the date is before the start date of game (15th January 2019) it will redirect to game() instead
         """
-
         if datetime.date.today() < datetime.date(2019, 1, 15):
             await self.game.callback(self, ctx)
             return
```
joker314: Two blank lines in a row will make Travis fail. When merging the two copies of `on_message`, don't forget to remove this stray newline to ensure Travis passes!

bot/cogs/cyber.py
maishams: omg thank u ily will do once the thingy finishes

bot/cogs/cyber.py
```diff
@@ -368,7 +370,16 @@ def suffix_from_number(num):
...
-
+        #Level Function Regex.
+        match = self.game_level_regex.match(message.content)
+        if match:
+            level_num, challenge_num = int(match.group(2)), int(match.group(3))
+            if match and message.channel.id == '532654287386443777':
+                base = 'hq'
+            elif match and message.channel.id == '532654358660120586':
+                base = 'm'
+            elif match and message.channel.id == '32654399453921291':
```
joker314: These channel IDs should ideally be placed in the constants file (src/bon/constants.py)

bot/cogs/cyber.py
```diff
@@ -368,6 +370,17 @@ def suffix_from_number(num):
             text += "We’re currently allocating students to their preferred locations so it’s an ongoing process!"
             text += " We’ll send out details of your location as soon as we can. It shouldn’t be too long!"
             await message.channel.send(text)
+        # Level Function Regex.
```
joker314: My personal preference would be to have an extra blank line before this. Your choice here.

```suggestion

        # Level Function Regex.
```

bot/cogs/cyber.py
```diff
@@ -78,11 +82,9 @@ def __init__(self, bot: Bot):
         Gets information about a specific CyberStart Game level and challenge.
         If the date is before the start date of game (15th January 2019) it will redirect to game() instead
         """
-
```
AnonGuy: Why were these blank lines removed?

bot/cogs/cyber.py
```diff
@@ -78,11 +82,9 @@ def __init__(self, bot: Bot):
...
         if datetime.date.today() < datetime.date(2019, 1, 15):
             await self.game.callback(self, ctx)
             return
-
```
AnonGuy: Why were these blank lines removed?

bot/cogs/cyber.py
```diff
@@ -368,6 +370,17 @@ def suffix_from_number(num):
             text += "We’re currently allocating students to their preferred locations so it’s an ongoing process!"
             text += " We’ll send out details of your location as soon as we can. It shouldn’t be too long!"
             await message.channel.send(text)
+        # Level Function Regex.
+        match = self.game_level_regex.match(message.content)
+        if match:
+            level_num, challenge_num = int(match.group(2)), int(match.group(3))
+            if match and message.channel.id == '532654287386443777':
```
joker314: Here, you do `if match and ...`, but you're already inside an `if` block which guarantees that `match` is defined, so it should be safe to remove the checks here.

bot/cogs/cyber.py
maishams: idk. ill put em back in

AnonGuy: cool and good

bot/cogs/cyber.py
```diff
@@ -368,6 +372,17 @@ def suffix_from_number(num):
             text += "We’re currently allocating students to their preferred locations so it’s an ongoing process!"
             text += " We’ll send out details of your location as soon as we can. It shouldn’t be too long!"
             await message.channel.send(text)
+        # Level Function Regex.
```
joker314: Consider adding a blank line here too?

```suggestion

        # Level Function Regex.
```

maishams: will do rn

bot/cogs/cyber.py
maishams: omds i should check all of these b4 committing. ty will do this rn!

maishams:
im going to sleep on this.

bot/cogs/cyber.py
```diff
@@ -369,6 +374,18 @@ def suffix_from_number(num):
             text += " We’ll send out details of your location as soon as we can. It shouldn’t be too long!"
             await message.channel.send(text)
 
+        # Level Function Regex.
+        match = self.game_level_regex.match(message.content)
+        if match and message.channel.id == HQ_ID or MOON_ID or FORENSICS_ID:
```
joker314: So this is parsed as

```python
if match and (message.channel.id == HQ_ID) or (MOON_ID) or (FORENSICS_ID):
```

Now, this will _always_ be true because you've set the variable `MOON_ID` and `FORENSICS_ID` to have non-empty value.

Instead, consider

```python
if match and message.channel.id in [HQ_ID, MOON_ID, FORENSICS_ID]:
```

maishams:
i got skiros to test my code bc mac is gay. it works
unknown-2

AlphexGaming:
can confirm works

bot/cogs/cyber.py
```diff
@@ -63,6 +63,10 @@ def __init__(self, bot: Bot):
             r"^.*\bwhat\b.*\belite\b.*\bemail\b.*$",
             re.IGNORECASE
         )
+        self.game_level_regex = re.compile(
+            r"^(?!:level).*(le?ve?l)\b.*?(\d{1,2}).*(\d{1,2})\b.*$",
```
joker314: You added a negative lookahead for "level".This should not be needed because `on_message` checks if a valid command has been included, and if so, ignores it. 

Is there evidence that this protection wasn't working? It's just that this method isn't great if we decide to change our prefix in future. :)

maishams: skiros tried the old code and it didnt work

fractalpenguin: You can add a check for the current command prefix using a format string, rather than hardcoding the command prefix.

joker314: True! But we really ought to investigate why the current checks are failing and if it affects the other regular expressions.

Further, :l (abbreviated) also triggers the command, but trying to code that into the regular expression is still a bad idea as cases could be missed and false-positives are likely to arise too.

fractalpenguin: the other regular expressions aren't using a negative lookahead, but it wouldn't hurt to check them again.

joker314: Okay. I'll investigate this evening. I'm okay with this being merged before then and we can fix later.

The negative lookahead can stay until then

joker314: @maishams I removed the extra check locally and nothing broke. When I typed `:level 1 13`, only one output showed (handled by the command). Without the colon or with text around it, also one output showed (handled by the regex as a fallback). I tested this locally on my own Discord server.

Could you check with skiros to make sure it didn't work. Maybe they just made a suggestion without quite testing it?

maishams: oh my old code didn't assign the variables right loool im an idiot

joker314: @maishams 
> oh my old code didn't assign the variable right

yes
> loool im an idiot

no 
<hr>
Ah alright. So if we all agree that the extra negative lookahead isn't needed, I think we should remove it to remove clutter in the code.

justanotherfailure:
Regex should be changed to this. Previously HQ 12 12 did not work, fixed by making 3rd catch group greedy.
`^(le?ve?l)\b.*?(\d{1,2}).*?(\d{1,2})\b.*$`

bot/constants.py
```diff
@@ -5,6 +5,9 @@
 QUOTES_CHANNEL_ID = 463657120441696256
 QUOTES_BOT_ID = 292953664492929025
 LOGGING_CHANNEL_ID = 538494690601992212
+HQ_ID = 532654287386443777
```
joker314: These are listed in the "Fun" constants section. Maybe they should have a comment at the top of them (and a blank line) to make it clear what their purpose is.

e.g.
```suggestion

# Channels for discussion about Game
HQ_ID = 532654287386443777
```

bot/cogs/cyber.py
```diff
@@ -63,6 +63,10 @@ def __init__(self, bot: Bot):
             r"^.*\bwhat\b.*\belite\b.*\bemail\b.*$",
             re.IGNORECASE
         )
+        self.game_level_regex = re.compile(
+            r"^(le?ve?l)\b.*?(\d{1,2}).*?(\d{1,2})\b.*$",
```
teamshortcut: At the moment (as I understand it) this regex will fire if the Level or Challenge number is out of range, which shouldn't be the case.

bot/cogs/cyber.py
```diff
@@ -369,6 +374,20 @@ def suffix_from_number(num):
             text += " We’ll send out details of your location as soon as we can. It shouldn’t be too long!"
             await message.channel.send(text)
 
+        # Level Function Regex.
+        match = self.game_level_regex.match(message.content)
+        if match and message.channel.id in [HQ_ID, MOON_ID, FORENSICS_ID]:
```
teamshortcut: This is great; could this be incorporated into the default behavior for the regular level command?

teamshortcut:
Oh, and the regex should probably catch alternate notation; so 13 1 (covered), 13.1, L13C1, l13c1, so on.

teamshortcut:
It may be helpful if the regex only fired if the message were along the lines of "can I have a hint on..." or "I need some help on..." to avoid it popping up too much and obstructing discussion.

bot/cogs/cyber.py
joker314: Maybe... but `:level x y` has the same problem.

justanotherfailure:
Change the regex to 
`^(le?ve?l)?\b.*?(\d{1,2}).*?(\d{1,2})\b.*$`
image

bot/cogs/cyber.py
justanotherfailure:
```suggestion
            r"^(le?ve?l)?\b.*?(\d{1,2}).*?(\d{1,2})\b.*$",
```

joker314: Two numbers in the message shouldn't really trigger the command, do you agree?

justanotherfailure: since this only runs in the relevant channels, it shouldnt be too much of an issue on misfires

however the regex can be edited to avoid them completely

justanotherfailure: I've updated the regex so at least an L is required in the message to prevent missfires

So the following will work (any case):

level 125
level125
level 125
level 12 5
lvl 125
l12c5
l13.1

maishams: thanks skiros!!

bot/cogs/cyber.py
AnonGuy: I can implement this in a separate PR.

bot/cogs/cyber.py
```diff
@@ -63,6 +63,10 @@ def __init__(self, bot: Bot):
             r"^.*\bwhat\b.*\belite\b.*\bemail\b.*$",
             re.IGNORECASE
         )
+        self.game_level_regex = re.compile(
+            r"^(le?v?e?l?).*?(\d{1,2}).*?(\d{1,2})\b.*$",
```
joker314: Okay, so this regular expression has changed a lot. It matches this, which I think it shouldn't:

> **L**ol, remember the **2018** intake?

`:level 20 18`

This would maybe be acceptable, but in addition, the command will report out-of-range errors back to the user, in the form of a spammy message. We need to suppress these error messages if the command was invoked via this pattern match rather than a direct invocation of the command.

iThe bot replies with "Invalid challenge number!" if asked for level 20 19

<hr>

I think, also, in https://github.com/CyberDiscovery/cyberdisc-bot/pull/135/commits/54517984e7b282bfe49ecf55ff9338dafdb8ede5 we accidentally removed the `.*` prefixing the pattern for "level". This is bad because you can no longer use inline phrases like

> I'm struggling with level 4.3

Maybe we ought to remove the `^` anchor at the beginning of the match, as that seems more readable than adding a `.*`.

Also, notice that "Lol" matches. This probably isn't a good idea. Maybe we should only match "L" on its own. "Lev" probably shouldn't be matched either, but "lvl" and "level" should be.

This change will resolve some of the above:

<hr>

Also, "2019" shouldn't match. There should be a digit-boundary between each number, so that "20 19" is okay, but three-digit numbers should not work.

Here is a regular expression to consolidate the above requests:

```suggestion
            r"(le?v?e?l|l(?=\b|\d))\D*(\d{1,2})\D+(\d{1,2})\b.*$",
```

Matches
- LxCy
- lvl x y
- lvl x.y
- levl x y
- lvel x y
- level x y
- lvl x challenge y

Does not match:
- lev x y
- love x y
- level xy
- level 01 04 (note leading zeros)

We should maybe consider allowing the last one.

<hr>

Additionally, you use a capturing group to match for "level". I haven't changed it here because it would require changing some of your other code, but you never use this result so maybe we should use a non-capturing group `(?:...)` so as to only match what we need.

<hr>

I really like how this has all turned out and am open to thoughts on the above!

bot/cogs/cyber.py
```diff
@@ -369,6 +374,20 @@ def suffix_from_number(num):
...
+        if match and message.channel.id in [HQ_ID, MOON_ID, FORENSICS_ID]:
+            level_num, challenge_num = int(match.group(2)), int(match.group(3))
+            if message.channel.id == HQ_ID:
+                base = 'hq'
+            elif message.channel.id == MOON_ID:
+                base = 'm'
+            elif message.channel.id == FORENSICS_ID:
+                base = 'f'
+            message.content = f"{self.bot.user.mention} level {base} {level_num} {challenge_num}"
+            new_ctx = await self.bot.get_context(message)
```
justanotherfailure:
```suggestion
            ctx.invoke(self.bot.get_command('level'), base, level_num, challenge_num)
```
slightly less hacky way of doing this.

bot/cogs/cyber.py
```diff
@@ -369,6 +374,20 @@ def suffix_from_number(num):
...
+            await self.bot.invoke(new_ctx)
```
justanotherfailure:
```suggestion
```
also get rid of this line of accepting the above suggestion.

bot/cogs/cyber.py
```diff
@@ -369,6 +374,20 @@ def suffix_from_number(num):
...
+        match = self.game_level_regex.match(message.content)
+        if match and message.channel.id in [HQ_ID, MOON_ID, FORENSICS_ID]:
+            level_num, challenge_num = int(match.group(2)), int(match.group(3))
+            if message.channel.id == HQ_ID:
+                base = 'hq'
+            elif message.channel.id == MOON_ID:
+                base = 'm'
+            elif message.channel.id == FORENSICS_ID:
+                base = 'f'
+            message.content = f"{self.bot.user.mention} level {base} {level_num} {challenge_num}"
```
justanotherfailure:
```suggestion
```

AnonGuy:
As a heads up @maishams @justanotherfailure, I'm going to be closing this soon if merge conflicts aren't resolved. The PR has been open for over a month.

PR #136, opened 2019-02-08
Implemented BytesIO for image generation

### LINKED ISSUE #134: Bot-generated images added to staging area
**Describe the bug**
The bot-generated images appear on git status.

**To Reproduce**
1. Use a create-text-image based command e.g ":jibhat"

**Expected behavior**
The image added should not be added to the staging area.

**Additional context**
This can be fixed by adding the generated images to the .gitignore

AnonGuy:
This PR just changes the image generation command so that it uses an `io.BytesIO` object to write (and send) the image, rather than saving it to disk. This is a better approach for several reasons:

- Commands that are executed asynchronously display **only the image that was requested**. With the previous approach, this was not guaranteed since these images are saved to disk with the same filename.
- There are no longer any files saved to disk, meaning issues with the staging area (such as #134) are resolved.
- The command is marginally faster, as it is no longer blocked by disk r/w speeds.

resolves #134

### DIFF
--- a/bot/cogs/fun.py
+++ b/bot/cogs/fun.py
@@ -2,6 +2,7 @@
 Set of bot commands designed for general leisure.
 """
 import textwrap
+from io import BytesIO
 from random import choice, randint
 from string import ascii_lowercase
 from typing import AsyncGenerator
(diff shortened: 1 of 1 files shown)


AnonGuy:
I have tested this feature locally: <br/>
image

PR #145, opened 2019-02-14
Replace Pipenv with Poetry

AnonGuy:
This PR replaces our `Pipfile` and `Pipfile.lock` with the `pyproject.toml` and `poetry.lock`.
Using poetry will decrease our build time significantly.


thebeanogamer:
I think this is ready, but it needs the draft tag removed to allow Travis to test it.

.travis.yml
```diff
@@ -12,11 +12,12 @@ services:
 install:
   - docker build -t cd-bot .
   - 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then docker run -d -e BOT_TOKEN --name cd-bot cd-bot; fi'
-  - pip install -r requirements-ci.txt
+  - pip install poetry
```
thebeanogamer: Why do we actually need to do this if Travis doesn't run the bot? Surely this just extends the build.

thebeanogamer:
giphy-8

PR #149, opened 2019-02-18
Setup Sentry 2.0

thebeanogamer:
Add Sentry.io logging.

I'd appreciate @AnonGuy looking at the lockfile as I'm not sure that poetry has changed it in the right way.

The environment variable has been set in the hosting environment.

### DIFF
--- a/cdbot/__init__.py
+++ b/cdbot/__init__.py
@@ -4,7 +4,11 @@ from os import environ
 
 from .bot import bot
 
+import sentry_sdk
+
 
 def main():
     """Entry point for poetry script."""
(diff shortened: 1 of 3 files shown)


jos-b:
I'm pretty sure this won't setup sentry like you want sentry to be setup. Discord.py has an error handler so I don't know if sentry ever gets access to the errors that are raised by the bot. Is this intentional?

jos-b:
I may be incorrect and sentry may report (please do tell me if it does) but I think [these docs](https://docs.sentry.io/error-reporting/capturing/?platform=python) can be used to report exceptions which occur in commands.

PR #150, opened 2019-02-19
Add poetry.lock to .gitignore

AnonGuy:


thebeanogamer:
I don't see how this solves the problem.

AnonGuy:
Adding the lockfile to .gitignore prevents anyone from accidentally adding dev-dependencies to it when making a PR. It's quite easy for someone to do `poetry install` without passing `--no-dev`, which would prolong our build time by quite a bit.

AnonGuy:
It doesn't remove the lockfile from the repo, just prevents changes being made to it unless they are made deliberately.

PR #153, opened 2019-02-25
[WIP] Fix speech bubble text size.

justanotherfailure:
Modified create text image command to make large lengths of text fit inside the speech bubble.

Edit: Kind of work in progress, some stuff isnt too good rn, but its getting there


justanotherfailure:
ill submit a new one when i actually figure out a not hacky hacky way to do this

PR #155, opened 2019-03-08
Bot reacts with 👋 to any welcome message

AlphexGaming:


cdbot/cogs/fun.py
```diff
@@ -151,6 +151,10 @@ def check(reaction, user):
         if "revive" in message.content.lower():
             await emojify(message, "nou")
 
+        # Adds waving emoji when a new user joins.
+        if "Welcome to the Cyber Discovery" in message.content and message.author.id == 155149108183695360:
```
joker314: Please create a constant in [the constants file](https://github.com/CyberDiscovery/cyberdisc-bot/blob/master/cdbot/constants.py) and then reference the bot ID there. This will require you to add the variable name you choose to the top of _this_ file so as to ensure it is imported.

The reasons for this are:
- If someone else is running our bot, and has chosen a different instance of the welcome-message bot, they don't need to dig through our code
- If we need to swap out the welcome-message bot, we again do not need to dig through the code to find what to change
- The significance of the number "155149108183695360" is obvious immediately, because it will have a descriptive variable name associated to it.

cdbot/cogs/fun.py
```diff
@@ -151,6 +151,10 @@ def check(reaction, user):
         if "revive" in message.content.lower():
             await emojify(message, "nou")
 
+        # Adds waving emoji when a new user joins.
+        if "Welcome to the Cyber Discovery" in message.content and message.author.id == WELCOME_BOT_ID
+            await message.add_reaction("👋")
```
joker314: So a later pull request has now been merged in, #157, which uses a new syntax for referencing emoji characters. Instead of including the literal character, which can be difficult to view in some text editors, we now refer to non-ASCII characters by *name*.

That is to say
```suggestion
            await message.add_reaction("\N{THUMBS UP SIGN}")
```
Could you make that change here so as to keep the codebase consistent?

AnonGuy: What jokebook said

PR #157, opened 2019-03-09
Emoji improvements

jos-b:
I've cleaned up the code relating to auto-reacting to things like `kali` and `duck`.

I also removed the usage of emoji constants like `🦆` with their unicode escapes, in this case it would be `"\N{DUCK}"` which is best practice to prevent encoding errors with contributors opening the file.

### DIFF
--- a/cdbot/cogs/fun.py
+++ b/cdbot/cogs/fun.py
@@ -29,11 +29,12 @@ REACT_TRIGGERS = {
     "revive": "nou"
 }
 
-def convert_emoji(message: str) -> typing.List[]:
+
+def convert_emoji(message: str) -> List[str]:
     """Convert a string to a list of emojis."""
     emoji_trans = list(map(iter, EMOJI_LETTERS))
     # Enumerate characters in the message
-    
+
     emojified = []
 
     for character in message:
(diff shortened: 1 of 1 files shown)


cdbot/cogs/fun.py
```diff
@@ -127,29 +140,23 @@ def check(reaction, user):
...
-        # React if message contains Kali.
-        if "kali" in message.content.lower():
-            await message.add_reaction("🚔")
-
-        # React if message contains Duck.
-        if "duck" in message.content.lower():
-            await message.add_reaction("🦆")
+        for trigger in REACT_TRIGGERS:
+            # Check if the message contains a trigger
+            if trigger in message.content.lower():
```
joker314: Would lowercasing outside of the loop be more efficient as we'd only need to do it once?

thebeanogamer:
Why do I have an email with a message from @joker314, but nothing actually here?

joker314:
@thebeanogamer because I prematurely filed a comment.

Here were my full thoughts which I am now unable to leave as a review due to the pull request being merged.

> My concern with this is that it assumes that if a string contains exactly one character, that character will be a literal emoji.
>
>The problem, however, is that a string of length one could be a single character that should be converted into a regional indicator. For example, I think it is conceivable that in future, one would want to add an `F` reaction. This is of course just one character in length, but it should still be run through `emojify`.
>
>Instead, I offer two possible alternative implementations:
> - At the top, in the dictionary literal, run the relevant values (in this case, `nou`) through a pure function which essentially maps `convert_emoji` over the string; *or*
> - Run *everything* through `emojify`, and give `emojify` the functionality that if it encounters an emoji, it does not attempt to convert it but instead runs `.add_reaction` directly on it.
>
>I especially like my second suggestion for another reason: I can envisage a situation where we want to react with *two* emoji. For example, a list of *all* of our thinking emojis for `dabato`. My understanding is that `.add_reaction` will only do one at a time.

jos-b:
@joker314 Did this get fixed?

Issue #147, opened 2019-02-15
Embeds for #readme


teamshortcut:
At the moment, #readme (in the server) is slightly hard to read, and spread across different messages. It would be nice if we could use the bot to resend this content with some nicer looking embeds. This would allow the content to be split up slightly more, and to have some fancy colours in there for readibility.

It would be heplful to have the content for this stored somewhere where admins are able to edit it, and some way (perhaps an admin-only) command to purge the current content of #readme and resend the updated text.

This text should include everything currently in there, perhaps slightly reformatted. This includes the rules of the server, clarifications, and explanations of our roles. This may also be a good opportunity to add explanations of a few of the channels, though this isn't exactly important.

AnonGuy:
If someone's going to implement this, we'll need a formatted and consistent set of rules, each with a different category for each embed.

thebeanogamer:
Resolved in #159

PR #159, opened 2019-03-10
Added the :readme command from scratch.

Sh3llcod3:
I've created a system which solves #147. Essentially it allows hot-plugging READMEs using a JSON config file. There is heavy validation on the command, and only `Root` role members will be able to use the command since it's a core server component.

This is my first PR to do with the bot, or to do with any Python of any sort so please keep in mind that I've never written Asyncio Python or used the discord.py library before.

This PR introduces the `:readme` command, which has two main functions.

* `:readme pull` will DM the requesting Root role member a copy of the readme JSON file, which they can edit at will. `get`, `fetch`, and others work too, please check the code to view all aliases.

* `:readme push <channel id> <interval in seconds, max 15 minutes>` will create the readme messages. **If the user uploads a JSON file with the command, it will attempt to use that** to create the messages, otherwise, if a JSON is not uploaded along with the command, it will use the default config as is on our server. Since a readme needs to be well spaced out, I've included the interval argument, which if optionally supplied will asynchronously wait before sending the next message for a clean message channel. Again there are many aliases and `push` isn't the only keyword.

As I've said before, the command features a lot of error-catching and validation, so only members of our server who have the `ROOT_ROLE_ID` constant in their roles will be allowed to use the command. 

The JSON syntax isn't too dissimilar to the way [discord embed visualiser works](https://leovoel.github.io/embed-visualizer/), and it can be used to create messages featuring text, embeds with fields and descriptions or a combination of the lot. One look through `cdbot/data/readme.json` should make it really obvious regarding the JSON syntax, but if you're not sure, you can ask me or use the [online JSON editor](https://jsoneditoronline.org/) to visualise the existing default JSON file. 

I would strongly recommend changing the default files `cdbot/data/readme.json` and `cdbot/data/readme_raw.json` if you intend to merge this PR and use it in production, as the config I've included is very basic and does not refactor the messages on our server as nicely-formatted embeds. I leave that task up to whoever is interested, as creativity is beyond my capabilities. 

I hope this PR has been useful. 

Thank you for taking the time to read this. 
With regards, Sh3llcod3


Sh3llcod3:
@thebeanogamer C'mon man. :woman_facepalming:

cdbot/cogs/cyber.py
```diff
@@ -155,6 +157,147 @@ def __init__(self, bot: Bot):
...
+        elif (channel_id == 0 and operand in README_SEND_ALIASES):
+            misssing_channel_embed = Embed(
+                colour=0xff5722,
+                description=":facepalm: **Whoops, you missed out the channel ID! Try again.**"
+            )
+            await ctx.message.delete()
+            await ctx.send(embed=misssing_channel_embed)
+
+        # Process the request.
+        elif ROOT_ROLE_ID in [msg_author.id for msg_author in ctx.message.author.roles]:
```
fractalpenguin: Instead of having an if-statement to check for the Root role, a command [check](https://discordpy.readthedocs.io/en/rewrite/ext/commands/api.html#discord.ext.commands.has_role) would be a better method of checking, as it hides the function when users type in `:help`

Sh3llcod3: I absolutely agree with you there, it would be a better form of control. I've just deleted my testing server environment, so it'll probably have to be sometime tomorrow or day after, for me to make this change, unless someone, could re-factor the conditional. Whichever is sooner :+1:

Sh3llcod3: @fractalpenguin I've refactored the conditional, squashed a bug and added the decorator. Does this look good? :+1:

Sh3llcod3:
@teamshortcut Yes, I have (non-exhaustively) tested the PR. It seems to work and I managed to fix a small bug. There are some caveats, that I think are worth outlining, __not that they impact functionality in a tangible way.__ :woman_shrugging: 

* It does not perform any form of JSON file validation. It assumes the JSON file uploaded will be valid and if its invalid, it will simply show an error in the chat and not display where the parse error happened exactly.
* There is no type checking. Just like all the other commands there is no checking of variable types. This shouldn't be a problem though, as none of the existing commands have any proper type-checking and supplying an invalid type argument will simply show an error in the log chat and not do anything.
* This one is again, something that won't really come into play much, but there is no message verification, so if Discord decides to randomly lose a message into the deep-blue there nothing that will detect this. The only reason this is here, since I believe I may have encountered a case of the wild disappearing message once :laughing: 

Either way, 
I've incorporated @fractalpenguin's changes, fixed bugs and tested it. Tweaks and writing a proper JSON config is all that remains I guess?

teamshortcut:
Looks great. I'd asked @lightspeedana to have a look at the embeds themselves, using her mystical art abilities, though if she's busy I can have a go at it.

PR #160, opened 2019-03-10
Refactor so as to expand possible autoreactions to single letters and multiple emojis

joker314:
Effectively, this pull request takes most of the logic from `emojify`, and puts into a new function `react_to_message`. Then, `emojify` takes its argument, lowercases it, `convert_emoji`s it, and then runs it through `react_to_message`.

This is because `emojify` is used very frequently in our codebase, yet we still need to use the logic that is now in `react_to_message`.

Instead of guessing whether a string is composed of an emoji or a letter by its length, convert letters into emoji first. This way, we can, in any case, run `react_to_message`.

This allows us to have automatic reactions constituting:
- One letter (such as 'Q')
- Multiple emoji (such as two different thinking faces)

<hr>

Testing was done, but not on the `:react` command as there were no other members in my testing server.

### DIFF
--- a/cdbot/cogs/fun.py
+++ b/cdbot/cogs/fun.py
@@ -6,7 +6,7 @@ import textwrap
 from io import BytesIO
 from random import randint
 from string import ascii_lowercase
-from typing import List
+from typing import Iterable, List
 from urllib.parse import urlencode
 
 import asyncpg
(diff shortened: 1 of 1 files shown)


Ge0rg3:
What do you mean by "Testing was done, but not on the :react command as there were no other members in my testing server."? How else could you test it on Discord?

joker314:
@Ge0rg3 The functionality is shared across multiple features of the server. For example, the bot automatically reacts to messages with certain content (like `revive`)

The `:react` command will react to a person other than the author of the message. Discord does not auto-populate servers with more than one person. As a result, testing that specific command was not possible. However, the `emojify` function can still be tested, as aforementioned, through other means.

Ge0rg3:
@joker314 iirc, you can still :react your own message?

jos-b:
Why was this closed - it looks like a genuine improvement?

joker314:
There's a bug that I haven't got around to investigating

fractalpenguin:
You can keep this PR open, since it is still relevant.

PR #166, opened 2019-03-14
Update Game autoresponder for new end date

thebeanogamer:


thebeanogamer:
@maishams Seems to have beat me to this by about 10 seconds so closing.

PR #172, opened 2019-03-24
More Kubernetes Fixes

thebeanogamer:
image


fractalpenguin:
The error that you described is as a result of not specifying the proper image name for the Docker container.

PR #175, opened 2019-03-30
Fixed quotes function and added former user quote fetching

fractalpenguin:
Fixes the `:quotes` function that has been broken since the last update to discord.py, and implements former user quote fetching via user ID.

### DIFF
--- a/cdbot/cogs/fun.py
+++ b/cdbot/cogs/fun.py
@@ -12,22 +12,23 @@ from urllib.parse import urlencode
 import asyncpg
 from aiohttp import ClientSession
 from cdbot.constants import (
-    ADMIN_ROLES, EMOJI_LETTERS, FAKE_ROLE_ID, PostgreSQL, QUOTES_BOT_ID, QUOTES_CHANNEL_ID, STAFF_ROLE_ID,
-    WELCOME_BOT_ID
+    ADMIN_ROLES,
+    EMOJI_LETTERS,
+    FAKE_ROLE_ID,
+    QUOTES_BOT_ID,
+    QUOTES_CHANNEL_ID,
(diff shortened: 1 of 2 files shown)


thebeanogamer:
image
Failing tests need to be resolved before merging.
https://dev.azure.com/cyberdiscovery/Cyber%20Discovery%20Discord%20Bot/_build/results?buildId=90&view=ms.vss-test-web.build-test-results-tab

thebeanogamer:
Also, @jos-b doesn't have approving reviews perms.

cdbot/cogs/fun.py
```diff
@@ -12,22 +12,23 @@
...
-from discord import Embed, File, Member, Message, NotFound
-from discord.ext.commands import Bot, Cog, Context, command, has_any_role
+from discord import Embed, File, Message, NotFound
+from discord.ext.commands import Bot, Cog, Context, UserConverter, command, has_any_role
 from discord.utils import find as discord_find
 from PIL import Image, ImageDraw, ImageFont
 
 ascii_lowercase += " !?$"
 
-REACT_TRIGGERS = {
```
thebeanogamer: Why are we splitting some sections over multiple lines and not this one?

fractalpenguin: It can be put in a single line without exceeding the 120 character line limit.

thebeanogamer: For readability and easy expansion, this would be better split over multiple lines

thebeanogamer: Will approve if changed.

fractalpenguin: Done. The line limit should be made lower.

PR #176, opened 2019-04-01
Fix images properly this time

thebeanogamer:
This should fix `:jibhat`, although I would double check, as I made this PR on my phone in bed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN apk add --update --no-cache build-base
 RUN apk add --update --no-cache libffi-dev
 RUN apk add --update --no-cache zlib-dev
 RUN apk add --update --no-cache jpeg-dev
-RUN apk add --update --no-cache freetype
+RUN apk add --update --no-cache libfreetype6-dev
 
 COPY . /app
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


justanotherfailure:
Maybe it was a bad idea to switch to Pillow after all.

thebeanogamer:
👀
Screenshot_20190401-080728

justanotherfailure:
Whooops.

Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apk add --update --no-cache build-base
 RUN apk add --update --no-cache libffi-dev
 RUN apk add --update --no-cache zlib-dev
 RUN apk add --update --no-cache jpeg-dev
-RUN apk add --update --no-cache freetype
+RUN apk add --update --no-cache libfreetype6-dev
```
AnonGuy:
```suggestion
RUN apk add --update --no-cache freetype-dev
```
[`libfreetype6-dev`](https://pkgs.alpinelinux.org/packages?name=libfreetype6-dev&branch=edge) is not a valid package in the Alpine registry.
Like I've mentioned several times before, [`freetype-dev`](https://pkgs.alpinelinux.org/packages?name=freetype-dev&branch=edge) is the correct package name.

PR #181, opened 2019-04-04
Exclude "command ... not found" errors from log.py

AnonGuy:


cdbot/log.py
```diff
@@ -32,9 +33,8 @@ def _level_to_color(self, level_number: int):
         return LEVEL_COLORS.get(level_number)
 
     def emit(self, record):
-        if not self.client.loop.is_running():
-            # The event loop is not running (discord is not connected) so
-            # do not send the message
+        if not self.client.loop.is_running() or re.match('Command ".*" is not found', str(record.msg)):
```
fractalpenguin: Would it not be better to check the error directly on the `on_command_error` handler, rather than using a regex query for the message?

PR #183, opened 2019-04-06
Added to fun.py

MJayEl:
Added a catpic command which gets random cat pictures and a fun fact command.


teamshortcut:
Although we appreciate the work you've put into this, the features are not relevant to the bot, and are covered by other bots used in the server (e.g. Dyno or Unbelieaboat) should we want to implement them.

PR #189, opened 2019-04-14
Nickname moderation

teamshortcut:
This pull request ~~should~~ add nickname moderation to the bot. It will, based on an array of regex patterns, check on nickname changes/removal and username changes (and member joins) if the display name hits any of the filters. (defined in constants.py) If so, it will remove the nickname, or when the username needs filtering it will assign a placeholder nickname.

I have tested this fully on my own bot, but not using the cogs or constants system. _basically this https://hastebin.com/axoradaqan.py version worked but the cogs system confuses me slightly so I'm not sure if this is entirely correct._ I can test this properly on a local version of the bot tomorrow (15/04/2019), or if anybody would like to before then, that'd be much appreciated. _I think last time I did this there was some kerfuffle over having to make other functions `async` which may or may not be an issue again._

The regex array can be updated after staff discussion, and currently just has a single basic check (for invite links) that I have tested.


cdbot/constants.py
```diff
@@ -54,6 +54,9 @@ class PostgreSQL:
     "Forensics": ["forensics", "f"],
 }
 
+# Admin Constants
+PLACEHOLDER_NICKNAME = "Valued server member"
+NICKNAME_PATTERNS = ["(discord.gg/)"]
```
joker314:
```suggestion
NICKNAME_PATTERNS = ["(discord\.gg/)"]
```
Dot matches any character (including but not limited to a literal dot). 

Though the more vague the regular expression the better so maybe this can be left as-is. (Ideally, we'd have a system where if the Levenchtein distance is 1 or 2 we'd still detect the nickname).

cdbot/cogs/admin.py
```diff
@@ -0,0 +1,55 @@
...
+            # assign placeholder nickname
+            await memberAfter.edit(nick=PLACEHOLDER_NICKNAME)
+
+
+    @Cog.listener()
+    async def on_member_join(member):
+        username = member.name
+        if checkName(username):  # bad username
+            # assign placeholder nickname
+            await member.edit(nick=PLACEHOLDER_NICKNAME)
```
joker314: Have you considered resetting the nickname entirely (to the username) instead of to a predefined constant? If so, what's the reasoning out of interest?

cdbot/cogs/admin.py
```diff
@@ -0,0 +1,55 @@
...
+from discord import Member
+from discord.ext.commands import Bot, Cog, Context, command
+
+from cbot.constants import (
+    PLACEHOLDER_NICKNAME, NICKNAME_PATTERNS
+)
+
+def checkName(nick):
+    result = False
+    for i in NICKNAME_PATTERNS:
```
joker314: This is a good way to do this but I'm thinking whether it would be more Pythonic to use a for/else construction? In this case I see no obvious advantage either way (except maybe some kind of minor efficiency improvement) but I'll leave it to the more experienced for further commentary.

cdbot/cogs/admin.py
picapi: I believe the code at this line is checking if the username of the user joining is suitable; setting a placeholder nickname for the user removes the offending username of the newly joined user. I believe the idea is to reset the nickname if the username is OK, otherwise set the nickname of the user to hide the username and remove the nickname.

fractalpenguin: That users could have inappropriate usernames, or usernames that cannot be typed on an English keyboard (due to using non-ASCII/non-Latin characters).

cdbot/cogs/admin.py
fractalpenguin: It could be replaced by a list comprehension:
``
return any(re.match(i,nick) for i in NICKNAME_PATTERNS)
``

cdbot/cogs/admin.py
fractalpenguin: Furthermore, if the username is fine, it will reset the nickname to the username.

joker314: Ah, yep, nevermind

thebeanogamer:
With your current spout of activity and Game coming to an end, should we expect to actually see you back on the [Discord server](https://discord.gg/Kf8n5rT) @joker314?

You seem to contribute a lot to the repos, but don't get to actually take advantage of any of the code. Whilst such selflessness is admirable, I do question its sense.

cdbot/cogs/admin.py
```diff
@@ -0,0 +1,51 @@
+import re
+
+from cbot.constants import NICKNAME_PATTERNS, PLACEHOLDER_NICKNAME
```
fractalpenguin:
```suggestion
from cdbot.constants import NICKNAME_PATTERNS, PLACEHOLDER_NICKNAME
```
Would cause an import error.

cdbot/cogs/admin.py
```diff
@@ -0,0 +1,51 @@
...
+from discord.ext.commands import Bot, Cog
+
+
+def checkName(nick):
```
fractalpenguin: Maybe replace with a list comprehension?
```suggestion
def checkName(nick):
    return any(re.match(i,nick) for i in NICKNAME_PATTERNS)
```

fractalpenguin: Or alternatively, a more efficient for loop:
```suggestion
def checkName(nick):
    for i in NICKNAME_PATTERNS:
         if re.match(i,nick):
             return True
    return False
```

joker314:
@thebeanogamer:

Jump to:
- <a href="#joke-beano">In response to Beano</a>
- <a href="#joke-pr">In response to this Pull Request</a>

<a name="joke-beano"></a>
# In response to Beano
I think about this a lot. I dislike the server immensely because it is designed for cheating. This view is reinforced by two persons enrolled in an A-level course at my school and participating in the CyberDiscovery competition. One of them (P) achieved 100%, the other (Q) did not. Q joked/stated that he would "need help" from P if he was to complete the *Game* round of the competition in time, but that he was nonetheless very capable. P did not call Q out on this.

I remember, specifically, there was one challenge, which I had just solved (completely by myself). There was a key idea that needed to be had; and then the construction of an attack payload. Q noted that P had told Q the key idea, but Q defended themselves by saying they then constructed the payload all by themselves. P is in the wrong in this situation too, they have very little to lose due to their remarkably high percentage and are yet bringing their friends along to Elite rather than the people that deserve it. This competition looks quite good on university applications for some people, and we've learnt from the Q&A session that there may be certifications at the end which are useful for jobs.

I remember Q asked P for a flag and P obliged (though Q claims to have solved the problem before receiving the flag).

I write all this to show that Q and P are willing to cheat. I also note that just because you get 100% does not indicate that you will not have a very high academic integrity.

Anyway, P (or *maybe* Q but I think my memory is correct here) looked at Q's computer. Q observed a DM to a member on the Discord with whom I have interacted. It requested a flag. I don't know whether a response was received but P could have DMed every single member of the server until they received a response.

<hr>

The server is also used for 'hints'. I think this is wrong because not everybody is a member of the server, and so this provides an unfair advantage to the members. It feels like cheating. It's difficult, because often receiving hints (or even in-depth lessons) can be an incredibly educational experience, but that has to be balanced with the immense harm of denying someone the opportunity to attend the Elite stage.

There is very little to the server other than a list of people who have done well but who are willing to give out hints or DM you the flags; and channels *dedicated* to hint-giving.

There's a general chatroom, but given how most of the conversations are unrelated to cyber security and can thus be hosted elsewhere.

<hr>

You are correct, though, that it is much harder to give out hints during the Essentials stage of the competition, so I must think about this carefully: but it feels a bit weird interacting with people that have cheated; or supporting a server which is *almost* designed for cheating.

I don't know what I'll do.

<a name="joke-pr"></a>
# In response to this pull request

All of this looks good. This link change is good.

To expand on my comment on checking for slight spelling deviations, it's a hard thing, of course. For the discord.gg thing, a distance of 2 might be acceptable, but for certain obscene words, 1 might be too much ('duck'). So were we to add some kind of error-acceptance, it would need to be different for each banned word. That would probably be best done with our own custom data structure to keep everything packaged up nicely.

kaovd:
@joker314 
tl;dr 
qu1371Ib3ral

AnonGuy:
It would have been a better idea to have this conversation in DMs, not the review page of an irrelevant Pull Request.

cdbot/cogs/admin.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Admin functionality
+    """
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @Cog.listener()  # triggered on new/removed nickname
+    async def on_member_update(memberBefore, memberAfter):
+        newNickname = memberAfter.display_name
```
AnonGuy: Surprised our linters didn't pick up on this, but could we replace `camelCase` with `snake_case` thanks

AnonGuy: This applies to functions as well as variables

teamshortcut: Oh sorry, can do.

cdbot/constants.py
AnonGuy: We could use [`difflib.SequenceMatcher`](https://docs.python.org/3/library/difflib.html#sequencematcher-objects)? It's not Levenshtein, but it does do similarity ratios.

AnonGuy: image

AnonGuy: Also the double backslash isn't needed in this case.

joker314:
@anonguy - if a username is bad, the nickname is set to "Valued server member" or whatever the constant is to mask the username. We might want to kick the member and then DM them automatically requesting they pick a better username; but I'm happy with the way this PR handles thinhs

joker314:
Was the idea of changing the floor loop into a list comprehension rejected? If so, what was the reasoning?

This is otherwise perfect for merging.

PR #190, opened 2019-04-15
Nickname 2 electric boogaloo

teamshortcut:


cdbot/constants.py
```diff
@@ -58,7 +59,11 @@ class PostgreSQL:
 
 # Admin Constants
 PLACEHOLDER_NICKNAME = "Valued server member"
-NICKNAME_PATTERNS = [r"(discord\.gg/)"]
+NICKNAME_PATTERNS = [
+    r'(discord\.gg/)',  # invite links
+    r'(nigg|cunt|ligma|fag|nazi|hitler|\bpaki\b)',  # banned words
+    r'(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)'  # hyperlinks
```
joker314: I don't like this regex. It matches a lot of things that aren't URLs.

The HTTPS is optional, so is the www, so unless I'm missing something it'll match "abc.def ghi". Unless someone forgets a space, I can't think of a valid use case but I still think this is a little lax.

Further, why are the /s escaped in the HTTPS bit? The delimiters in this regex are not slashes and they aren't otherwise reserved. This probably doesn't matter because there's no harm in overescaping punctuation but we might as well remove the backslashes?

PR #191, opened 2019-04-15
Case insensitivity

teamshortcut:


teamshortcut:
_whoops sorrry forgot this_

PR #203, opened 2019-04-19
Speed up lint

thebeanogamer:
(Hopefully)

This might not work and I'll close it if it doesn't.


thebeanogamer:
It does work! 🙌

thebeanogamer:
This reduces the lint time from 1m 24s to 41s by removing `python-setuptools` as the Azure agents already have Python installed.

PR #202, opened 2019-04-19
Add Stackdriver Debugger Support

thebeanogamer:
Enabled [Stackdriver Debugger](https://cloud.google.com/debugger/) to help debug errors.

PR #204, opened 2019-04-20
Suppress some errors

jos-b:
This adds some code so that errors which don't need reporting are not reported, but are instead given feedback from the bot in the channel at time of execution.

This surpasses:
- Command not found

Provides feedback and suppression for:
- Missing required arguments
- Check failures
- Cooldowns

Adds feedback but does not suppress:
- Argument parsing errors
- Bad argument errors
- All other exceptions

### DIFF
--- a/cdbot/cogs/general.py
+++ b/cdbot/cogs/general.py
@@ -55,7 +55,6 @@ class General(Cog):
         else:
             await ctx.send("\N{NO ENTRY SIGN} An error occured during execution, the error has been reported.")
 
-
         self.bot.log.exception(error)
 
 
--- a/cdbot/cogs/general.py
+++ b/cdbot/cogs/general.py
@@ -32,10 +32,10 @@ class General(Cog):
     @Cog.listener()
     async def on_command_error(self, ctx, error):
         # Try provide some user feedback instead of logging all errors.
-        # 
+
         if isinstance(error, commands.CommandNotFound):
             return  # No need to log unfound commands anywhere or return feedback
-        
+
         if isinstance(error, commands.MissingRequiredArgument):
             # Missing arguments are likely human error so do not need logging
             parameter_name = error.param.name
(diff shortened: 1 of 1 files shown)


thebeanogamer:
Will this affect Sentry and Google Stackdriver or just `#developer-log`?

jos-b:
Do they work by adding logging handlers or are they exception-hooks based?

thebeanogamer:
I think:tm: Sentry is a logging handler and Stackdriver is exception based.

thebeanogamer:
But I'd check that

jos-b:
I don't see the issue with suppressing any of the human error things from Sentry anyway

jos-b:
We should be trying to find code errors, not typos and wrong argument usage.

jos-b:
I can't even find any code referencing google stack driver, where is it used?

thebeanogamer:
No, it's not an issue at all. Anything to reduce my email spam is a good thing, and I'll be going through some of the common errors at some point to add validation.

thebeanogamer:
Stackdriver is added by Google Kubernetes Engine, although #202 will improve its output.

thebeanogamer:
Send me an email if you want access to it, I have to add you to the Cloud Console.

jos-b:
From what I can tell both do use logging handlers, so the errors where I've removed the logging will mean the exceptions go from both solutions. Not a hard thing to change, it's as simple as removing a `return`. It's up to you whether we need to keep in the CommandNotFound & Check failures etc.

thebeanogamer:
If you're suppressing it from the log, ideally it would be surpressed from Sentry and Stackdriver as well.

jos-b:
[Stackdriver docs only reference a logging handler](https://cloud.google.com/logging/docs/setup/python#installing_the_library) but Sentry's new SDK makes it unclear how everything works.

jos-b:
The old Sentry SDK (raven) made you use logging handlers but looking [here](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/hub.py) it looks like exception hooks are used instead.

jos-b:
Though [here](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/logging.py) it looks like the implementation is logging. No clear documentation from Sentry on this one.

thebeanogamer:
Urgh, is it going to request a review from both members of the team?

thebeanogamer:
Oh no, it just needs me to review again because the team wasn't working when I reviewed it the first time.

jos-b:
Yeah, it requires an approving review from at least one member of the team AIUI.

jos-b:
Error

PR #206, opened 2019-04-20
Update GitHub README

thebeanogamer:
Update `README.md` because it's horrifically out of date.


README.md
```diff
@@ -34,21 +34,25 @@ Finally, run the command `poetry run cdbot` in the root of the repository to run
 * **`:help`** - Displays information about the usage and syntax of the commands
 
 ### Admin Commands
-* **`:set_quote_channel [#channel]`** - Set the channel to be used as a source by the `:quotes` command. Note: You will have to do this after every restart of the bot.
-* **`:mute [@mention]`** - Mute the selected user indefinitely.
-* **`:unnmute [@mention]`** - Unmute the selected user.
+* **`:migrate_quotes`** - Migrate the existing quotes in the set quotes channel to the PostgreSQL database.
+* **`:readme [push | pull] [channel id] [interval]`** - `pull` will DM the user a copy of the JSON used for #readme. `push` will create the readme channel using the set JSON file.
```
teamshortcut: I would have said `:readme` is better left out of readme.md, as it's a Root only command. Same for `:migrate_quotes`, they clutter the information in `:help` and can't be used by the vast majority of users anyway.

joker314: I believe the help command only shows commands which the invoking user has the power to use.

This is separate to this README which in my view should be a guide to developers (who will need to work on all aspects of the bot or at least understand them)

thebeanogamer: Surely the point of the Admin Commands section is to include things like this?

teamshortcut: Yes, though I would argue that we don't need an Admin Commands section at all. If @joker314 is correct though, I have no problem with it.

thebeanogamer: @joker314 is correct, `:help` only shows commands the user can run. I would also argue that it would be bad practice to include commands that aren't in the documentation.

PR #208, opened 2019-04-20
General Repo Improvements

thebeanogamer:
* Removed `.dockerignore`
* Added an emoji reaction template
* Fixed codeowners


.github/ISSUE_TEMPLATE/emoji_request.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+**Reaction Trigger**
+What text should cause the reaction?
+
+**Reaction Emoji**
+What emojis should be reacted to the message?
+
+**Explanation**
+Why should we add this? What in-joke is this a reference to that no one else will get?
```
picapi: I understand the intent behind the question; but I feel like the "What in-joke" remark may have a slightly unwelcoming tone to it - depending on the intent of the remark it may be served better in my opinion by either asking "What does this reaction reference?" or "Do note that references only a couple of people understand may not be suitable".

thebeanogamer: That seems much better and I shall change it.

PR #210, opened 2019-04-20
Reaction improvements

### LINKED ISSUE #199: Significant lag during automatic reactions
Because we aren't cancelling any of the automatic reactions once one's been triggered, it's possible to cause considerable lag for the bot by just sending a few messages:

image

jos-b:
Add some limits to auto-reactions to stop the bot from lagging to high heaven.

Closes #199

### DIFF
--- a/cdbot/cogs/fun.py
+++ b/cdbot/cogs/fun.py
@@ -37,6 +37,14 @@ REACT_TRIGGERS = {
     "kali": "\N{ONCOMING POLICE CAR}",
     "duck": "\N{DUCK}",
     "revive": "no u",
+    "child": "fbi open up",
+    "fbi": "fbi open up",
+    "loli": "fbi open up",
+    "hentai": "fbi open up",
+    "illegal": "fbi open up",
+    "maltego": "fbi open up",
+    "@here": "who pinged",
+    "@everyone": "who pinged"
 }
 
 
@@ -173,16 +181,6 @@ class Fun(Cog):
         if message.content.startswith(":react ") or message.content.startswith(":emojify "):
             # Don't react to invocations of :react
             return
-
-        # React if a message contains an @here or @everyone mention.
-        if any(mention in message.content for mention in ("@here", "@everyone")):
-            await message.add_reaction("\N{SLIGHTLY FROWNING FACE}")
-            return await emojify(message, "who pinged")
-
-        # React FBI OPEN UP if message contains trigger words.
-        triggers = ["child", "fbi", "loli", "hentai", "illegal", "maltego"]
-        if any(trigger in message.content.lower() for trigger in triggers):
-            return await emojify(message, "fbi open up")
 
         for word in message.content.lower().split():
             # Check if the message contains a trigger
--- a/cdbot/cogs/fun.py
+++ b/cdbot/cogs/fun.py
@@ -166,7 +166,7 @@ class Fun(Cog):
             finally:
                 await sent.delete()
                 
-        if message.content.startswith(":react "):
+        if message.content.startswith(":react ") or message.content.startswith(":emojify "):
             # Don't react to invocations of :react
             return
 
--- a/cdbot/cogs/fun.py
+++ b/cdbot/cogs/fun.py
@@ -177,7 +177,7 @@ class Fun(Cog):
                     await message.delete()
             finally:
                 await sent.delete()
-                
+
         if message.content.startswith(":react ") or message.content.startswith(":emojify "):
             # Don't react to invocations of :react
             return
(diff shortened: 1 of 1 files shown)


cdbot/cogs/fun.py
```diff
@@ -32,7 +32,7 @@
     "dabato": "\N{THINKING FACE}",
     "kali": "\N{ONCOMING POLICE CAR}",
     "duck": "\N{DUCK}",
-    "revive": "nou",
+    "revive": "no u",
```
thebeanogamer: Why is this reaction formatted differently to the ones above it?

jos-b: I added the space by mistake, will remove. It is because the current reaction is `NOU` but this would make it `NO U` with a spacing emoji.

thebeanogamer: I get that, but I was wondering why it isn't `\N{NOU}` like the ones above it.

jos-b: Ah, because the `\N` is an emoji escape. For example, `\N{DUCK}` is the literal :duck: emoji.

jos-b: no u is a text response so it gets converted into emojis at a later stage.

thebeanogamer: 👍 Makes sense

cdbot/cogs/fun.py
```diff
@@ -165,26 +165,32 @@ def check(reaction, user):
...
         if any(mention in message.content for mention in ("@here", "@everyone")):
             await message.add_reaction("\N{SLIGHTLY FROWNING FACE}")
-            await emojify(message, "who pinged")
+            return await emojify(message, "who pinged")
 
         # React FBI OPEN UP if message contains trigger words.
         triggers = ["child", "fbi", "loli", "hentai", "illegal", "maltego"]
         if any(trigger in message.content.lower() for trigger in triggers):
-            await emojify(message, "fbi open up")
+            return await emojify(message, "fbi open up")
```
thebeanogamer: Would it be worth moving this into the collection of other reactions on lines 31-36?

jos-b: Yeah, I guess, just wondering if it would be nice having so many duplicate responses in the dictionary. I don't think it's a huge issue though, will move after adding the ratelimit .

thebeanogamer:
See, the problem there is that you're using emacs.

jos-b:
I win once again 
image

cdbot/cogs/fun.py
thebeanogamer: :eyes:

cdbot/cogs/fun.py
```diff
@@ -166,26 +178,22 @@ def check(reaction, user):
...
-        # React if a message contains an @here or @everyone mention.
-        if any(mention in message.content for mention in ("@here", "@everyone")):
-            await message.add_reaction("\N{SLIGHTLY FROWNING FACE}")
-            await emojify(message, "who pinged")
-
-        # React FBI OPEN UP if message contains trigger words.
-        triggers = ["child", "fbi", "loli", "hentai", "illegal", "maltego"]
-        if any(trigger in message.content.lower() for trigger in triggers):
-            await emojify(message, "fbi open up")
+        if message.content.startswith(":react ") or message.content.startswith(":emojify "):
```
thebeanogamer: This seems like bad practice. Could you not just have it check if it's a valid command like this?
```suggestion
        if !ctx.valid:
```

cdbot/cogs/fun.py
```diff
@@ -166,26 +178,22 @@ def check(reaction, user):
...
-        if any(mention in message.content for mention in ("@here", "@everyone")):
-            await message.add_reaction("\N{SLIGHTLY FROWNING FACE}")
-            await emojify(message, "who pinged")
-
-        # React FBI OPEN UP if message contains trigger words.
-        triggers = ["child", "fbi", "loli", "hentai", "illegal", "maltego"]
-        if any(trigger in message.content.lower() for trigger in triggers):
-            await emojify(message, "fbi open up")
+        if ctx.valid:
+            # Don't react to invocations of :react
```
thebeanogamer:
```suggestion
            # Don't react to valid commands
```

thebeanogamer:
Well that didn't work as well as I'd hoped it would.
image

thebeanogamer:
(Using this as reference https://github.com/CyberDiscovery/cyberdisc-bot/blob/e4ceaadfc3db09fcfc19291da8be091bc74196bc/cdbot/cogs/cyber.py#L466)

jos-b:
Oh of course, it's in `on_message`

jos-b:
We could iterate through all bot prefixes?

jos-b:
Ah wait, now I see the full code snippet, will add now

jos-b:
I'm not sure this received adequate reviews, can we revert and retry?

PR #212, opened 2019-04-20
Add extra context

### LINKED ISSUE #211: Logging ideas
Currently the #developer-log is pretty vague. It gives a location for the erring code but does not tell us any metadata about invocation in the context of commands.

Since Python's `logging` module has [custom field](https://stackoverflow.com/a/17558757) support, we could change the Discord logging handler to make use of the metadata passed in, we could add:
- Command name
- Message ID
- Channel ID
- User mention

This would make it a lot easier to debug errors directly from the embed posted in #developer-log.

jos-b:
Adds extra context to logs sent in #developer-log.

Closes #211


jos-b:
Can I get some context for this PR?

PR #214, opened 2019-04-20
Round the cooldown retry after

jos-b:
This PR just rounds the cooldown durations so that they aren't a horrifically long float.

Prevents this:


jos-b:
♻️

PR #215, opened 2019-04-20
Jump to message log

jos-b:
Adds another field to the error embed which allows a message to be jumped to if the error occurred in a server.


jos-b:
you can't approve when it failed lint that's agains the law

jos-b:
beans there were so many mistakes with that previous commit I'm worried that you approved that...

AnonGuy:
image

jos-b:
than k

jos-b:
Jump 🆙 to :two: message 📥 logs 🌲

PR #217, opened 2019-04-21
Remove Have I Been Pwned Commands

jos-b:
No one uses them.

Bad idea.


jos-b:
We should add this back, you have been pwned.

thebeanogamer:
No one ever used this

PR #218, opened 2019-04-21
Switch Dockerfile to Debian

thebeanogamer:
Only merge if it improves build times.

Alpine builds took around 2m 42s.


thebeanogamer:
Yeah, I have no idea what these package names should be.

Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM python:3.7-alpine3.9
+FROM python:3.7-stretch
 
-RUN apk add build-base freetype-dev git jpeg-dev libffi-dev zlib-dev
+RUN apt update
+RUN apt install -y build-base freetype-dev git jpeg-dev libffi-dev zlib-dev
```
xxskiros: This should be `RUN apt install -y build-essential libfreetype6-dev git libjpeg-dev libffi-dev zlib1g`.

thebeanogamer: Thanks

xxskiros:
Builds seem to be around 30 seconds faster, switch to Debian I guess(?)

thebeanogamer:
Yeah, this reduces the build from 2m 42s to 2m 24s.

My only other thought, and @AnonGuy or @jos-b probably know the answer to this, is regarding the packages we are actually installing. `build-essentials` is the only one that isn't already installed. Could we remove it to speed up the build? It's currently taking up 2 of the longest steps in the build.
image

PR #219, opened 2019-04-21
Update requirements

jos-b:
- Update requirements
- Relock poetry.lock
- Remove extra discord.py step from Dockerfile

Bon Appétit


jos-b:
I think a bug was introduced in this PR, we should revert to try diagnose!

thebeanogamer:
What bug was introduced?

jos-b:
@thebeanogamer cyber discovery bot

PR #221, opened 2019-04-23
Add command to count 100% users

Bottersnike:
Fairly self explanatory. `('...' \n '...')` was used for string wrapping; lmk if project standards are a different format.


cdbot/cogs/cyber.py
```diff
@@ -346,6 +346,23 @@ def __init__(self, bot: Bot):
...
+        """
+        Gets the number of 100% and true 100% users
+        """
+
+        game = true = 0
+        for member in ctx.guild.members:
+            for role in member.role:
+                if role.name == '100 Percent':
+                    game += 1
+                elif role.name == 'True 100%':
```
fractalpenguin: By using `elif` instead of a separate `if` statement, you count members with True 100% separately. More importantly, every member with the True 100% role will also have the 100% role, meaning the output for True 100% will always be 0.

Bottersnike: That's actually not true because it iterates over all their roles. The _is_ a problem though in that snippet that my shourly disapear form a push.

fractalpenguin: Nevermind.

cdbot/cogs/cyber.py
```diff
@@ -346,6 +346,19 @@ def __init__(self, bot: Bot):
...
         await self.countdown('5th March 2019', 'CyberStart Essentials', ctx)
 
+    @command()
+    async def hundered(self, ctx: Context):
+        """
+        Gets the number of 100% and true 100% users
+        """
+
+        game_r = utils.get(ctx.guild.roles, name='100 Percent')
+        true_r = utils.get(ctx.guild.roles, name='True 100%')
```
thebeanogamer: This should be using an id stored in `constants.py`, rather than a hard-coded name. This is both bad practice and could potentially break if role names change or there are multiple roles with the same name, as has been known to happen.

thebeanogamer: True 100%: 552536443948367903
100%: 532654673270669313

cdbot/cogs/cyber.py
```diff
@@ -346,6 +346,19 @@ def __init__(self, bot: Bot):
 
         await self.countdown('5th March 2019', 'CyberStart Essentials', ctx)
 
+    @command()
+    async def hundered(self, ctx: Context):
```
AnonGuy:
```suggestion
    async def hundred(self, ctx: Context):
```

cdbot/cogs/cyber.py
```diff
@@ -346,6 +347,18 @@ def __init__(self, bot: Bot):
...
+        """
+        Gets the number of 100% and true 100% users
+        """
+
+        game_r = utils.get(ctx.guild.roles, id=HUNDERED_ROLE_ID)
+        true_r = utils.get(ctx.guild.roles, id=TRUE_HUNDERED_ROLE_ID)
+
+        await ctx.send((f"There are {len(game_r.members)} 100% users, of which {len(true_r.members)} have also "
```
AnonGuy: Maybe this could be more specific?
```suggestion
        await ctx.send((f"There are {len(game_r.members)} users who completed game, of which {len(true_r.members)} have also "
```

cdbot/cogs/cyber.py
Bottersnike: Fixed

cdbot/cogs/cyber.py
```diff
@@ -353,10 +353,10 @@ def __init__(self, bot: Bot):
         Gets the number of 100% and true 100% users
         """
 
-        game_r = utils.get(ctx.guild.roles, id=HUNDRED_ROLE_ID)
-        true_r = utils.get(ctx.guild.roles, id=TRUE_HUNDRED_ROLE_ID)
+        game_r = utils.get(ctx.guild.roles, id=HUNDRED_PERCENT_ROLE_ID)
+        true_r = utils.get(ctx.guild.roles, id=TRUE_HUNDRED_PERCENT_ROLE_ID)
 
-        await ctx.send((f"There are {len(game_r.members)} 100% users, of which {len(true_r.members)} have also "
+        await ctx.send((f"There are {len(game_r.members)} that have completed Cyberstart Game. Out of them, {len(true_r.members)} have also "
```
Bottersnike: This wraps over the 120-char that was being used to wrap the original line. Probably worth re-wrapping.

cdbot/cogs/cyber.py
```diff
@@ -346,6 +347,18 @@ def __init__(self, bot: Bot):
...
+    @command()
+    async def hundred(self, ctx: Context):
+        """
+        Gets the number of 100% and true 100% users
+        """
+
+        game_r = ctx.guild.get_role(HUNDRED_PERCENT_ROLE_ID)
+        true_r = ctx.guild.get_role(TRUE_HUNDRED_PERCENT_ROLE_ID)
+
+        await ctx.send((f"There are {len(game_r.members)} that have completed Cyberstart Game. Out of them, {len(true_r.members)} have also "
```
AnonGuy: Not super important, but what's the extra set of parens for?

fractalpenguin: Line wrapping.

Bottersnike: It allows for string continuation without a `+`. Thinking about it, python actually allows that with the inital set of parens for the function call, so they're just redundant :D

fractalpenguin: See PEP8 guidelines: https://www.python.org/dev/peps/pep-0008/#id19

AnonGuy: yes

AnonGuy: @fractalpenguin Since when was an extra set of parens required for implicit string concatenation?

fractalpenguin: ¯\_(ツ)_/¯

AnonGuy: I don't recall anything in PEP8 about this.

PR #222, opened 2019-04-26
Update regex for past tense

thebeanogamer:
DO NOT MERGE UNTIL CYBERSTART ESSENTIALS HAS CLOSED


joker314:
We ought to automate this in future, checking the centrally stored dates. Maybe a Word or Sentence class if we're being ambitious. This should also be where we handle pluralisation

thebeanogamer:
Reviews now please @teamshortcut @AnonGuy

PR #225, opened 2019-05-07
Readme update

teamshortcut:
Updates Mentor list


Ge0rg3:
I feel like this isn't the cleanest way to do this. Don't we have a config file for this type of thing?

fractalpenguin:
@Ge0rg3, this is the config file.

fractalpenguin:
Also, R.I.P @AnonGuy

teamshortcut:
@Ge0rg3 you're absolutely right, but that's deliberate on my part. Linus has very kindly included the ability to upload a json file when running the command, any Root or Sudo can make a new readme that way without having to edit the default file. As a matter of policy though, I've encouraged staff to do it this way, so it has to be approved before sending and we can track the changes over time.

Ge0rg3:
I feel like we'd be better off storing the users separately, but we can do that some other time maybe :)

thebeanogamer:
Yeah @Ge0rg3, whilst we might look at that down the line, this is more of a short term fix to remove @AnonGuy as he requested.

PR #232, opened 2019-06-06
Add more Elite stats: young, old, undisclosed

joker314:
This adds some more statistics to the `:elitecount` command. The data points added are

#### Total young

This is the number of people who are going to Birmingham, Lancaster, or London in the 14-15 ("younger") age range. This does *not* take the symmetric difference of all three sets: instead, it just adds together the numbers for the three venues. This means that if somebody has the role for "Birmingham" and the role for "Lancaster" at the same time, they will be double-counted.

Could the server staff advise if this is every something which is done -- if so, perhaps taking the symmetric difference would be better.

#### Total old

Same as **total young**, but counts the people who have qualified for the "older" age group.

#### Undisclosed

This is the number of people who have the "Elite 2019" role, but do not have a role denoting the specific venue. This is useful information because it allows people to construct an upper and lower bound for each venue at any particular point in time

<hr>

This pull request is untested and therefore is being opened initially as a draft. However, given that the automated test suite seems to not error, I am upgrading this pull request to a proper one.

### DIFF
--- a/cdbot/cogs/cyber.py
+++ b/cdbot/cogs/cyber.py
@@ -377,7 +377,11 @@ class Cyber(Cog):
         exch_l_r = ctx.guild.get_role(ELITE_EXCH_LIST_ROLE_ID)
         exch_c_r = ctx.guild.get_role(ELITE_EXCH_CONF_ROLE_ID)
 
-        await ctx.send(f"""There are {len(elite_r.members)} server members that have qualified for CyberStart Elite.
+        elite_count = len(elite_r.members)
+        young_count = len(ldn_y_r.members) + len(brm_y_r.members) + len(lan_y_r.members)
+        old_count = len(ldn_o_r.members) + len(brm_o_r.members) + len(lan_o_r.members)
+
+        await ctx.send(f"""There are {elite_count} server members that have qualified for CyberStart Elite.
 {len(exch_l_r.members)} members have qualified for Exchange ({len(exch_c_r.members)} confirmed).
 
 Of those who didn't, preferences have been expressed as follows:
@@ -387,6 +391,11 @@ Birmingham - Younger: {len(brm_y_r.members)}
 Birmingham - Older: {len(brm_o_r.members)}
 Lancaster - Younger: {len(lan_y_r.members)}
 Lancaster - Older: {len(lan_o_r.members)}
+
+Total younger: {young_count}
+Total older: {old_count}
+
+There are {elite_count - (young_count + old_count)} server members who qualified but haven't told us their preferences
         """)
 
     async def countdown(self, countdown_target_str: str, stage_name: str, ctx: Context):


cdbot/cogs/cyber.py
```diff
@@ -387,6 +391,11 @@ def __init__(self, bot: Bot):
 Birmingham - Older: {len(brm_o_r.members)}
 Lancaster - Younger: {len(lan_y_r.members)}
 Lancaster - Older: {len(lan_o_r.members)}
+
+Total younger: {young_count}
+Total older: {old_count}
+
+There are {elite_count - (young_count + old_count)} server members who qualified but haven't told us their preferences
```
teamshortcut:
```suggestion
There are {elite_count - (young_count + old_count)} server members who qualified but haven't told us their preferences.
```

teamshortcut: This unduly irritates me.

thebeanogamer:
I'm probably going to rewrite the version of this currently deployed to be a bit more compliant with Python standards, so I won’t really talk about the code itself, I’m more concerned regarding the actual function of these changes. Whilst I am more than open to alternatives, in my mind, the point of this function is to allow access to information which is normally only available through administration tools, and this modification doesn’t do that. It manipulates available data rather than making more data available, in a manner in which users are able to do themselves if they are that desperate for the data.

Additionally, the output for this command is already very big. This output adds a further 5 lines to a command that is already big enough for GitHub to collapse the middle of it when the start and end are modified. I’m just not convinced that the ratio of users who want this information and are not willing to calculate it themselves to the number who will be annoyed by an even larger command spam is worthwhile.

Furthermore, we’d ask that you test PRs before submitting them (I am aware that I’m guilty of failing to do this). The automated tests do not check that the code works as intended, simply that it passes linting checks and that the Docker container can be built. The full configuration for the CI checks can be found here.

https://github.com/joker314/cyberdisc-bot/blob/1bb91c197bfc881bb044f31e34195c3f3faf979a/cdbot/cogs/cyber.py#L398

Finally, I take issue with this line. Whilst there are obviously some users who know which camp they are attending but have chosen to not tell us, either due to their desire for privacy or because they simply don’t care about the role, there are also users who lack the role because they are unable to attend a camp, or have not yet decided. To rewrite this to cover all possible eventualities would be bloated and wasteful, and I’m just not convinced that it would be constructive.

thebeanogamer:
I don't know why that line's not unfurling, but it's this:
```python
There are {elite_count - (young_count + old_count)} server members who qualified but haven't told us their preferences```

joker314:
Yeah, that makes sense 👍 

The line isn't unfurling because you clicked on the line number and copied the updated URL of the page. You should've instead copied a permalink to the line (which can be done by clicking the three dots which appear upon hovering near the line number).

thebeanogamer:
I copied the permalink? I followed the guide here because I wasn't sure why it wasn't working. 🤷‍♂ doesn't really matter.

joker314:
Ah, never mind, so you did. Hmmph.

Your link in your last comment, by the way, was treated as a link to an image -- which prevented me from opening it.

PR #237, opened 2019-06-15
Add :elitecountdown command

Bentechy66:
The command works fairly simply, just use `:elitecountdown <Location> <Age Group>` and it will respond with a countdown. Simples!

### DIFF
--- a/cdbot/cogs/cyber.py
+++ b/cdbot/cogs/cyber.py
@@ -366,7 +366,16 @@ class Cyber(Cog):
         Gets the time until elite, and shows in an embed
         """
         valid_elite_locations = ["London", "Birmingham", "Lancaster "]
-        elite_aliases = {}
+        elite_aliases = {
+            "Lo": "London",
+            "La": "Lancaster",
+            "B": "Birmingham",
+            "Brum": "Birmingham",
+            "Birm": "Birmingham",
+            "Lanc": "Lancaster",
+            "Lon": "London"
+        }
+
         # Validate and normalise inputs
         event_name = event_name.lower().capitalize()
         if event_name:
--- a/cdbot/cogs/cyber.py
+++ b/cdbot/cogs/cyber.py
@@ -369,15 +369,14 @@ class Cyber(Cog):
         # Validate and normalise inputs
         if event_name:
             event_name = event_name.capitalize()
-            if not (event_name in ELITE_LOCATIONS):
-                try:
-                    event_name = ELITE_ALISES[event_name]
-                except KeyError:
-                    content = (
-                        f"{event_name} is not a valid elite location! "
-                        f"Please choose from {', '.join(ELITE_LOCATIONS)}"
-                    )
-                    return await ctx.send(content)
+            try:
+                event_name = ELITE_ALISES[event_name]
+            except KeyError:
+                content = (
+                    f"{event_name} is not a valid elite location! "
+                    f"Please choose from {', '.join(ELITE_LOCATIONS)}"
+                )
+                return await ctx.send(content)
 
         if event_level:
             event_level = event_level[0].lower()
--- a/cdbot/cogs/cyber.py
+++ b/cdbot/cogs/cyber.py
@@ -9,8 +9,8 @@ from json import load
 
 from aiohttp import ClientSession
 from cdbot.constants import (
-    BASE_ALIASES, CYBERDISC_ICON_URL, END_README_MESSAGE, HINTS_LIMIT, HUNDRED_PERCENT_ROLE_ID, ROOT_ROLE_ID, Roles,
-    TRUE_HUNDRED_PERCENT_ROLE_ID, ELITE_ALISES, ELITE_DATE_KEYS
+    BASE_ALIASES, CYBERDISC_ICON_URL, ELITE_ALISES, ELITE_DATE_KEYS, END_README_MESSAGE, HINTS_LIMIT,
+    HUNDRED_PERCENT_ROLE_ID, ROOT_ROLE_ID, Roles, TRUE_HUNDRED_PERCENT_ROLE_ID
 )
 from dateutil.parser import parse
 from dateutil.relativedelta import relativedelta
@@ -366,7 +366,6 @@ class Cyber(Cog):
         Gets the time until elite, and shows in an embed
         """
         valid_elite_locations = ["London", "Birmingham", "Lancaster "]
-
 
         # Validate and normalise inputs
         event_name = event_name.capitalize()
--- a/cdbot/cogs/cyber.py
+++ b/cdbot/cogs/cyber.py
@@ -361,6 +361,101 @@ class Cyber(Cog):
                        f"{len(true_r.members)} have also completed Essentials and Assess.")
 
     @command()
+    async def elitecountdown(self, ctx: Context, event_name: str=None, event_level: str=None):
+        """
+        Gets the time until elite, and shows in an embed
+        """
+        valid_elite_locations = ["London", "Birmingham", "Lancaster "]
+        elite_aliases = {}
+        # Validate and normalise inputs
+        event_name = event_name.lower().capitalize()
+        if event_name:
+            if not (event_name in valid_elite_locations):
+                try:
+                    event_name = elite_aliases[event_name]
+                except:
+                    await ctx.send(f"{event_name} is not a valid elite location! Please choose from 'London', 'Birmingham', 'Lancaster'")
+                    return
+
+        if event_level:
+            event_level = event_level[0].lower()
+            if not (event_level in ["o", "y"]):
+                await ctx.send(f"Sorry, but that isn't a valid age group! Please choose from 'older', 'younger'")
+                return
+
+        # Make array of dates to get
+
+        events_to_get = []
+
+        if event_name:
+            if event_level:
+                events_to_get.append(event_name + event_level)
+            else:
+                events_to_get.append(event_name + "o")
+                events_to_get.append(event_name + "y")
+        else:
+            for location in valid_elite_locations:
+                events_to_get.append(location + "o")
+                events_to_get.append(location + "y")
+
+        # Get dates and send
+
+        event_date_keys = {
+            "Birminghamy": "23rd July 2019",
+            "Lancastery": "30th July 2019",
+            "Londony": "6th August 2019",
+            "Birminghamo": "22nd July 2019",
+            "Lancastero": "29th July 2019",
+            "Londono": "5th August 2019"
+        }
+        events = []
+        for event in events_to_get:
+            countdown_target = parse(event_date_keys[event]).date()
+
+            # Get the current date
+            today = datetime.date.today()
+            time_until_target = relativedelta(countdown_target, today)
+
+            # Given a number of items, determine whether it should be pluralised.
+            # Then, return the suffix of 's' if it should be, and '' if it shouldn't.
+            def suffix_from_number(num):
+                return "" if num == 1 else "s"
+
+            month_or_months = "month" + suffix_from_number(time_until_target.months)
+            day_or_days = "day" + suffix_from_number(time_until_target.days)
+
+            month_countdown = f"{time_until_target.months} {month_or_months}"
+            day_countdown = f"{time_until_target.days} {day_or_days}"
+
+            # Diable the months component of the countdown when there are no months left
+            if time_until_target.months:
+                month_and_day_countdown = f"{month_countdown} and {day_countdown}"
+            else:
+                month_and_day_countdown = day_countdown
+
+            if today > countdown_target:
+                events.append([event, "begun"])
+            events.append([event, month_and_day_countdown])
+
+        # Display to user
+
+        if len(events) == 1:
+            # No need for an embed
+            await ctx.send(f"The {events[0][:-1]} event for the {'older' if events[0][-1:] == 'o' else 'younger'} age group starts in {events[0][1]}!")
+            return
+        else:
+            embed = Embed(title="Cyberstart Elite: Camp Countdown", description="**Countdown:**\n", colour=Colour(0xae444a))
+
+            embed.set_thumbnail(url=CYBERDISC_ICON_URL)
+
+            for event in events:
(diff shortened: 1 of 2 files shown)


NilsIrl:
You could squash your commits

teamshortcut:
Has this been tested and shown to actually work? I also have some concerns about the quality of the code; readability is good in theory, but 100 lines for a countdown function seems excessive, to say the least. The approach also seems annoying to maintain; you have hardcoded values `+"o"`/ `+"y"` to allow the use of what I presume are meant to be jokey keys in a dictionary. It's confusing to people unfamiliar with the codebase, makes it harder to maintain in future and seems bad practice in general. I also personally question the necessity of a dictionary for this at all, but I also haven't spent the 10 seconds of thought required to come up with a solution so in the absence of any better solution, if it actually works I have less of a problem with this in isolation.

joker314:
Further to @teamshortcut's comments, I think you should avoid the `+"o"`/`+"y"` by having a nested dictionary. Take a look at how a nested dictionary is utilised over here:

https://github.com/CyberDiscovery/cyberdisc-bot/blob/90730351b734fd822a09f8a089b299df5e17cd94/cdbot/cogs/cyber.py#L369-L382

You could use your alias code to map "o" and "y" into "Older" and "Younger" too.

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,74 @@ def __init__(self, bot: Bot):
...
+
+        # Validate and normalise inputs
+        event_name = event_name.capitalize()
+        if event_name:
+            if not (event_name in valid_elite_locations):
+                try:
+                    event_name = ELITE_ALISES[event_name]
+                except KeyError:
+                    content = f"{event_name} is not a valid elite location! " \
+                        "Please choose from 'London', 'Birmingham', 'Lancaster'"
```
NilsIrl: You could have non hard-coded event locations.

NilsIrl: You could have non hard-coded event locations.

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,74 @@ def __init__(self, bot: Bot):
...
+    @command()
+    async def elitecountdown(self, ctx: Context, event_name: str = None, event_level: str = None):
+        """
+        Gets the time until elite, and shows in an embed
+        """
+        valid_elite_locations = ["London", "Birmingham", "Lancaster "]
+
+        # Validate and normalise inputs
+        event_name = event_name.capitalize()
+        if event_name:
```
NilsIrl: This line is useless

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,74 @@ def __init__(self, bot: Bot):
...
+            if not (event_name in valid_elite_locations):
```
NilsIrl: This as well cause you could have
```python
"London": "London"
```
in `ELITE_ALISES`

cdbot/cogs/cyber.py
NilsIrl: You could have non hard-coded event locations.

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,74 @@ def __init__(self, bot: Bot):
...
+                except KeyError:
+                    content = f"{event_name} is not a valid elite location! " \
+                        "Please choose from 'London', 'Birmingham', 'Lancaster'"
+                    await ctx.send(content)
+                    return
+
+        if event_level:
+            event_level = event_level[0].lower()
+            if not (event_level in ["o", "y"]):
+                await ctx.send(f"Sorry, but that isn't a valid age group! Please choose from 'older', 'younger'")
```
NilsIrl: Why are you using an `f"` string here?

cdbot/cogs/cyber.py
joker314: Specifically, you already have a constant which stores all of the valid locations, so you could just do something like

```python
f"Please chose from {[f"'{location}'" for location in valid_elite_locations].join(", ")}"
```

NilsIrl: Do you mean
```python
f"Please chose from {", ".join(valid_elite_locations)}"
```

NilsIrl: Or whatever you kids do these days to format a string inside a string (`f"`, `%` or `.format()`)

joker314: Sure, assuming you don't want single quotes around the items.

Bentechy66:
Next push will be dev, don't expect it to work, will update when working

joker314:
I'm thinking: what if we merged the dictionaries used in the elitecount command with the one in this PR. We could have an even deeper nesting, like a `role` key and a `date` key. Thoughts?

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,73 @@ def __init__(self, bot: Bot):
...
+        """
+
+        # Validate and normalise inputs
+        if event_name:
+            event_name = event_name.capitalize()
+            if not (event_name in ELITE_LOCATIONS):
+                try:
+                    event_name = ELITE_ALISES[event_name]
+                except KeyError:
+                    content = f"{event_name} is not a valid elite location! " \
```
jos-b: Backslash syntax is not pretty. I do not like it. It makes me sad.

```py
content = (
    f"{event_name}..."
    f"blah"
)
```

I like this.

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,73 @@ def __init__(self, bot: Bot):
...
+                        f"Please choose from {', '.join(ELITE_LOCATIONS)}"
+                    await ctx.send(content)
```
jos-b: This does not need two lines, just do `return await ctx.send(content)`

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,73 @@ def __init__(self, bot: Bot):
...
+                    return
+
+        if event_level:
+            event_level = event_level[0].lower()
+            if not (event_level in ["o", "y"]):
+                await ctx.send(f"Sorry, but that isn't a valid age group! Please choose from 'older', 'younger'")
```
jos-b: Again, why two lines, `return await ...`

cdbot/cogs/cyber.py
Bentechy66: incorrect, `event_name` can be `None`

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,73 @@ def __init__(self, bot: Bot):
...
+
+        if len(events) == 1:
+            # No need for an embed
+            await ctx.send(f"The {events[0][:-1]} event for the {'older' if events[0][-1:] == 'o' else 'younger'} "
+                           f"age group starts in {events[0][1]}!")
+            return
+        else:
+            embed = Embed(title="Cyberstart Elite: Camp Countdown",
+                          description="**Countdown:**\n",
+                          colour=Colour(0xae444a))
```
jos-b: Let's make this a constant.

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,73 @@ def __init__(self, bot: Bot):
...
+
+            embed.set_thumbnail(url=CYBERDISC_ICON_URL)
+
+            for event in events:
+                embed.add_field(name=f"{event[:-1]} - {'older' if events[0][-1:] == 'o' else 'younger'}",
```
jos-b: No. Don't do this inside an f-string. I don't like this.

cdbot/cogs/cyber.py
```diff
@@ -437,7 +507,13 @@ def suffix_from_number(num):
             month_and_day_countdown = day_countdown
 
         if today > countdown_target:
-            await ctx.send(f"{stage_name} has begun!")
+            return "Started"
```
jos-b: Add a newline after this if statement, it makes me sad.

cdbot/cogs/cyber.py
```diff
@@ -437,7 +507,13 @@ def suffix_from_number(num):
...
+        return month_and_day_countdown
+
+    async def countdown(self, countdown_target_str: str, stage_name: str, ctx: Context):
+        month_and_day_countdown = self.get_days_and_months_until(countdown_target_str)
+        if month_and_day_countdown == "Started":
+            await ctx.send(f"{stage_name} has already started!")
```
jos-b: Two lines can be one. `return await ...`

cdbot/constants.py
```diff
@@ -91,6 +91,36 @@ class Exchange:
     r'(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)'  # hyperlinks
 ]
 
+ELITE_ALISES = {
```
jos-b: This feels backwards. Why not have the camps with the aliases and like
```py
{
    "London": ["Lo", "Lon"]
}
```

Have some nice method for fetching them that way.

joker314: I agree in principle, but that is an O(N^2) solution. What we really need is an automated program/function to convert your pretty way of noting it down to a an efficient hash-table at the beginning of execution.

cdbot/cogs/cyber.py
NilsIrl: Even if it's None it won't change anything.

cdbot/constants.py
NilsIrl: > I agree in principle, but that is an O(N^2) solution

> Using python
> Talking about scalability

cdbot/cogs/cyber.py
Bentechy66: Personally I just think that is more confusing?

NilsIrl: Ask others but IMHO it's better and removes a useless check

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,73 @@ def __init__(self, bot: Bot):
...
+    @command()
+    async def elitecountdown(self, ctx: Context, event_name: str = None, event_level: str = None):
+        """
+        Gets the time until elite, and shows in an embed
+        """
+
+        # Validate and normalise inputs
+        if event_name:
+            event_name = event_name.capitalize()
+            if not (event_name in ELITE_LOCATIONS):
```
NilsIrl: I really believe this line should be removed in favour of a bigger `ELITE_ALISES` dictionnary.
Look at this code from another bot:
https://github.com/python-discord/seasonalbot/blob/9daf007c4094b8ee09b4779f78ac9122d0ca4e7d/bot/seasons/pride/pride_avatar.py#L13

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,72 @@ def __init__(self, bot: Bot):
...
+
+            embed.set_thumbnail(url=CYBERDISC_ICON_URL)
+
+            for event in events:
+                embed.add_field(name=f"{event[0]} - {event[1]}",
+                                value=f"Starting in {event[2]}",
+                                inline=True)
+
+            ctx.send(embed=embed)
+            return
```
thebeanogamer: These lines should be combined and awaited.

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,72 @@ def __init__(self, bot: Bot):
         await ctx.send(f"There are {len(game_r.members)} that have completed Cyberstart Game. Out of them, "
                        f"{len(true_r.members)} have also completed Essentials and Assess.")
 
+    @command()
+    async def elitecountdown(self, ctx: Context, event_name: str = None, event_level: str = None):
+        """
+        Gets the time until elite, and shows in an embed
+        """
```
thebeanogamer: The syntax for this command is never displayed. There should be a response to a command like this explaining how to use the command:
image

cdbot/cogs/cyber.py
```diff
@@ -360,6 +360,72 @@ def __init__(self, bot: Bot):
...
+
+        # Get dates and send
+
+        events = []
+        for event in events_to_get:
+            events.append([event[0], event[1], self.get_days_and_months_until(ELITE_DATE_KEYS[event[0]][event[1]])])
+
+        # Display to user
+
+        if len(events) == 1:
```
thebeanogamer: This doesn't work. This command got no reply:
image

cdbot/constants.py
```diff
@@ -91,6 +91,41 @@ class Exchange:
...
```
thebeanogamer: No please, make it stop. This makes me cry, and I'm the person who wrote this:
https://github.com/CyberDiscovery/cyberdisc-bot/blob/cd9eb426f288e57b19821e02d827fc12dc22c30c/cdbot/cogs/cyber.py#L364-L390

This can be massivly simplified in the way @jos-b documented.

cdbot/cogs/cyber.py
Bentechy66: `:elitecountdown` on its own will list all events

cdbot/constants.py
Bentechy66: I agree in principle, but that is an O(N^2) solution. What we really need is an automated program/function to convert your pretty way of noting it down to a an efficient hash-table at the beginning of execution.

cdbot/cogs/cyber.py
thebeanogamer: _Ummm, this isn't resolved?_

AnonGuy:
soon:tm:

Sh3llcod3:
Err, wouldn't it be better to have this function send an embed containing the countdown dates for all the camps and let the people lookup the dates for their camp, a bit like how the `:elitecount` command already works? Let's have the dates fixed in `Constants.py` and derive the countdown from that? :+1:

cdbot/constants.py
jos-b: hello just noticed typo this should be ALIASES not ALISES

thebeanogamer:
Given that this is getting very stale, doesn't work, and won't be needed in two weeks, this PR is being closed.

PR #239, opened 2019-07-25
Add brackets to emoji

clubby789:
↪️ and ↩️


Bottersnike:
```py
"some sting"
"something else"
```
evaluates to `"some stringsomething else"` when commas are missed. This has happened here.

Bottersnike:
https://github.com/CyberDiscovery/cyberdisc-bot/blob/2c778d39800972f237f36040a722f3918528c6b9/cdbot/cogs/fun.py#L33
will also require the addition of `()` to the end of the `+=` so that the parser is actually informed of the presence of the two new characters.

PR #257, opened 2019-10-01
Disable :elitecount

thebeanogamer:


Bottersnike:
Version control and config files be like

thebeanogamer:
I resent the suggestion that there are a million better ways to do this, at least one of which I will swap this too before merging

PR #260, opened 2019-10-06
Require Sudo/Root verification for official ping

AnonGuy:
Now that we have plenty of Sudos, and considering the abuse of the ping feature recently, I think it's worth requiring staff verification for official staff pings.

### DIFF
--- a/cdbot/cogs/fun.py
+++ b/cdbot/cogs/fun.py
@@ -149,7 +149,7 @@ class Fun(Cog):
             except asyncio.TimeoutError:
                 pass
             else:
-                if str(reaction) == "\N{THUMBS UP SIGN}" and user.top_role.name in ADMIN_ROLES:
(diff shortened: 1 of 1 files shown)


thebeanogamer:
Has this been tested with people other than an admin clicking the thumbs up button before an admin does?

thebeanogamer:
Also, I'm debating if pinging server staff would be a good idea. Feels like stuff will get missed if not, and people will spam server staff pings to ask for approvals anyway.

lightspeedana:
Looks good, but people may start spamming us with pings in order to verify. At the same time, I don't think there's much else that could be done so unless someone has an alternative, I'm fine with this 🤷‍♀

thebeanogamer:
~Not like Official Staff read their pings anyway~~

AnonGuy:
Any updates on this?

thebeanogamer:
#319 solved this

PR #267, opened 2019-10-31
add essential keyword

S10MC2015:
https://github.com/CyberDiscovery/meta/issues/290


Sh3llcod3:
@S10MC2015 Have you tested the regex locally?

ZomBMage:
The regex `essentials?` matches both "essential" and "essentials", because the question mark means the "s" is optional

thebeanogamer:
Closed as Stale

PR #277, opened 2019-11-06
Assess began/ended. Game began

S10MC2015:


thebeanogamer:
This is already handled by #270

S10MC2015:
Ah sorry

PR #291, opened 2019-11-20
Update all the things

thebeanogamer:
Update most packages in both `requirements-ci.txt` and `pyproject.toml`.

I've retired `requirements.txt` in an effort to standardise, but it remains to be seen what impact this will have on build times. Additionally, I can't see a good way of moving the CI pipeline to Poetry without forcing it to wait and install the dependencies for running the bot, Poetry has a `--no-dev`, but not a `--only-dev`.

This should also fix the fact that the bot is crashing every 30 seconds because of a Discord API change...


thebeanogamer:
Update: this increases container build time by about 20 seconds.

Sh3llcod3:
Ok so poetry is a mature enough build system. You're still using caret versioning so major versions will not update.

Sh3llcod3:
I have no idea whether the CI/Deployment could cope with installing based on the lockfile and having asterisk dependency versions. That way you update major versions with a single command

PR #306, opened 2019-12-31
Add Volcano challenges to :l

Bentechy66:


SSAFuze:
Redaction looks good, so have no objections either

PR #346, opened 2020-03-10
Block :react in #quotes

Bottersnike:
This should block `:react` in #quotes, which stops things like the attached image where people add random reaction messages to things in #quotes. ~~The theory with silently failing is that anyone who tries `:react` realises it fails, and can clean up their own message without needing a response from the bot cleaned up too.~~ Mentioned to Tref the other day, who suggested just diving in and making a PR. ~~Magic number used for the channel ID because I'm not aware of a quick and easy global configuration method.~~

image

### DIFF
--- a/cdbot/cogs/fun.py
+++ b/cdbot/cogs/fun.py
@@ -247,6 +247,9 @@ class Fun(Cog):
         Emojifies a given string, and reacts to a previous message
         with those emojis.
         """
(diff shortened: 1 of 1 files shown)


Bottersnike:
@thebeanogamer can do. What sort of delay would work? 10 seconds?

thebeanogamer:
10 seconds seems ok

Issue #353, opened 2020-03-10
Alllow voting off of quotes created prior to a bot restart


Bentechy66:
I believe this to be the cause, although embeds may also be the issue?

thebeanogamer:
This appears to be an issue with Discord.py caching. Raw_reaction_add might do what we need https://discordpy.readthedocs.io/en/latest/api.html#discord.on_raw_reaction_add?

thebeanogamer:
Excitingly, these errors appear to appear in Stackdriver but not bot logging:

```shell
discord.errors.NotFound: 404 Not Found (error code: 10008): Unknown Message
        at request (/usr/local/lib/python3.7/site-packages/discord/http.py:220)
        at delete (/usr/local/lib/python3.7/site-packages/discord/message.py:700)
        at on_reaction_add (/app/cdbot/cogs/fun.py:231)
        at _run_event (/usr/local/lib/python3.7/site-packages/discord/client.py:270)
```

thebeanogamer:
Resolved in #358

PR #357, opened 2020-03-11
Alllow voting off of quotes created prior to a bot restart

fractalpenguin:
Solves #353 by using `on_raw_reaction_add` rather than `on_reaction_add`, since the former is not dependent on the internal cache.


thebeanogamer:
Has this been tested?

fractalpenguin:
Please hold.

PR #358, opened 2020-03-11
Alllow voting off of quotes created prior to a bot restart

fractalpenguin:
Solves #353 by using on_raw_reaction_add rather than on_reaction_add, since the former is not dependent on the internal cache.

Turns out you cannot change a PR to a draft one, so the previous PR has been closed.


thebeanogamer:
Will this work if it's made to pass tests?

thebeanogamer:
Labelling is failing because of when this fork was created, it will be bypassed when this is merged.

Sh3llcod3:
*The ghost linter strikes again?*

thebeanogamer:
Merging with failing PR label is fine

Sh3llcod3:
PR author: *Hopefully this should work*
Linter: *Not so fast*

PR #382, opened 2020-04-15
Reimplement Quoting System

Skir0s:
does stuff, currently doesnt save stuff
implementation is a little bad

runs of a single base command quote
to save a quote, the first argument must be save
ie. `:quote save`
then any other arguments can follow
including `from` and `range`

`from` quotes from a specific channel
`range` quotes a range of messages

ok have a good day


thebeanogamer:
I'll look at getting the Mongo stuff setup when I get some time, also documentation in README please.

thebeanogamer:
Is this expecting the database to persist between reboots?

xxskiros:
> Is this expecting the database to persist between reboots?

yes, quotes cant be reverse engineered from their embeds, they dont contain enough data

thebeanogamer:
There are two ways to run Mongo within K8. There is the dodgy way which is just ["Run a Docker container"](https://github.com/GoogleCloudPlatform/mongodb-docker/blob/master/3/README.md#using-kubernetes) and the [proper way](https://kubernetes.io/blog/2017/01/running-mongodb-on-kubernetes-with-statefulsets/). I'd like to do the proper way, but my God is it complicated. This is probably also a good opportunity to change the deployment pipeline to the rolling update version we designed for Elections Bot.

thebeanogamer:
Ok I have spent the last 30 minutes trying to get my head round the proper way of running Mongo and I am completly lost. Let's do it the easy way...

thebeanogamer:
Right, that _should_ setup and run a persistent Mongo database.

thebeanogamer:
Also, we need to make a decision about `docker-compose.yml`

Sh3llcod3:
@xxskiros Is it ready to merge? Have you tested this?

xxskiros:
Very much no, pending @thebeanogamer

thebeanogamer:
Does this just need testing from my side?

xxskiros:
yes

Sh3llcod3:
I can't seem to update the branch due to a merge conflict, should be minor, but noted.

Skir0s:
Well, this might be a problem
image

thebeanogamer:
10/10

Skir0s:
The problem is caused by multiquotes having multiple owners, we could exclude them from the count however

Skir0s:
This is basically finished, just need to add some image supprt, but otherwise this can be considered ready to merge (after code review.)

cdbot/cogs/quoting.py
```diff
@@ -0,0 +1,382 @@
...
+from discord.ext import commands
+from discord.ext.commands import ArgumentParsingError, Bot, CheckFailure, Cog, UserConverter
+from motor.motor_asyncio import AsyncIOMotorClient
+
+from cdbot.constants import (
+    CYBERDISC_ICON_URL, LOGGING_CHANNEL_ID, MongoDB, QUOTES_CHANNEL_ID, QUOTES_DELETION_QUOTA, QUOTE_CZAR_ID, SERVER_ID
+)
+
+
+def is_quote_czar() -> Callable:
```
teamshortcut, line 14: I'd have thought this should be something more like `user_can_quote()`, as it checks for Quote Czar (which could be renamed) and admin?

cdbot/cogs/quoting.py
```diff
@@ -0,0 +1,382 @@
...
+        await ctx.send(f"{type(error)}: {error}")
+
+    @Cog.listener()
+    async def on_raw_reaction_add(self, raw_reaction: RawReactionActionEvent):
+        thumbs_down = "\N{THUMBS DOWN SIGN}"
+        if str(raw_reaction.emoji) == thumbs_down and raw_reaction.channel_id == QUOTES_CHANNEL_ID:
+            message = await self.quote_channel.fetch_message(raw_reaction.message_id)
+            reaction = [react for react in message.reactions if str(react.emoji) == thumbs_down][0]
+            if reaction.count > QUOTES_DELETION_QUOTA:
+                await self.database.delete_one({"_id": message.id})
```
teamshortcut, line 56: Consider moving this part to a function? I guess if it's not repeated it's not a massive deal but this seems like  it's just `delete_quote()`, and it'd make it easier to see what's happening here at a glance.

cdbot/cogs/quoting.py
```diff
@@ -0,0 +1,382 @@
...
+        return multi_quote
+
+    async def multi_quote_embed(self, multi_quote: dict) -> Embed:
+        embeds = [message.get("embed", None) for message in multi_quote["messages"]]
+        if any(embeds):
+            return list(filter(lambda x: x is not None, embeds))
+        embed = Embed().set_author(name=str(self.bot.user), icon_url=self.bot.user.avatar_url)
+        quoter = await self.get_member_by_id(multi_quote["quoted_by"])
+        channel = self.bot.get_channel(multi_quote["channel"])
+        footer_text = "Quoted by "
```
teamshortcut, line 158: This functionality is the same as L101 onwards (well, onwards a bit) it seems? Can it be moved to a function?

cdbot/cogs/quoting.py
```diff
@@ -0,0 +1,382 @@
...
+            original = await self.get_member_by_id(message["author"]["id"])
+            messages.append(
+                dict(content=message["content"], user=str(original) if not None else message["author"]["name"])
+            )
+        lines = []
+        for message in messages:
+            line_str = message["user"] + ": "
+            if "```" in message["content"]:
+                line_str += "\n"
+            line_str += message["content"][:1950]
```
teamshortcut, line 184: Will this always be sufficient, or can it truncate the content?

cdbot/cogs/quoting.py
```diff
@@ -0,0 +1,382 @@
...
+    async def generate_single_quote(self, quoted: Message, original: Message):
+        quote = self.quote_dict(quoted, original)
+        embed = await self.quote_embed(quote)
+        return embed, quote
+
+    async def generate_multi_quote(
+        self, channel: TextChannel, from_message: Message, to_message: Message, message: Message
+    ):
+        messages = await self.range_quote(channel, from_message, to_message)
+        if len("".join(message.content for message in messages)) > 2000:
```
teamshortcut, line 198: Is there no better way of handling this case? Can it not be split into multiple consecutive messages, as I'm assuming the 2000 comes from Discord's character limit?

cdbot/cogs/quoting.py
```diff
@@ -0,0 +1,382 @@
...
+            return await ctx.invoke(self.quote_save, message)
+        embed, _ = await self.generate_single_quote(message, ctx.message)
+        return await ctx.send(embed=embed)
+
+    @quote.group(name="from", invoke_without_command=True)
+    async def quote_from(self, ctx: commands.Context, channel: TextChannel, message: int):
+        """
+        Quote a message from another channel.
+        """
+        if not channel.permissions_for(ctx.author).read_messages:
```
teamshortcut, line 226: Nice catch.

cdbot/cogs/quoting.py
Skir0s: This is true, it can be done but I think simple annotations will help.

H4ckerJ4cker:
Thanks shorty

cdbot/cogs/quoting.py
Skir0s: Yeah, I'll rename it so it's clearer

cdbot/cogs/quoting.py
Skir0s: A lot of it is similar, but due to the differences in how multiquotes and normal quotes are stored, it'd be a bit more difficult.

cdbot/cogs/quoting.py
Skir0s: This will always cut short messages which are too long, it could however error if there are multiple very long lines so I will change how this is done, thanks for pointing this one out

cdbot/cogs/quoting.py
Skir0s: Thank you )

cdbot/cogs/quoting.py
```diff
@@ -179,9 +179,14 @@ async def multi_quote_embed(self, multi_quote: dict) -> Embed:
         lines = []
         for message in messages:
             line_str = message["user"] + ": "
-            if "```" in message["content"]:
+            if message["content"].startswith("```"):
                 line_str += "\n"
-            line_str += message["content"][:1950]
+            line_str += message["content"]
+            if len(line_str) > 200:
```
teamshortcut, line 185: Why is this 200?

Skir0s: At a maximum length of 200, there can be around 10 messages that can fit into a multiquote so I think that'd be a reasonable length to truncate to.

teamshortcut: Sure. Perhaps that should be moved a constant somewhere then? Is there a check somewhere for the length of the total combined content?

cdbot/cogs/quoting.py
Skir0s: At some point in this PR's history, I had included a paginator to make multiple embeds for very long quotes, however, the code was quite messy and so was removed. It can however be added at a later date or now if it's really needed with better code, but I removed purely because the chance of that occuing should be relatively low.

Skir0s: Actually, looking at the code, this check can probably be removed, more changed to make it so that it maxes out at 10 quotes rather than having an actual character limit.

cdbot/cogs/quoting.py
Skir0s: Yeah I'll add some comments for this in the code, no real need for a function here as you've said.

cdbot/cogs/quoting.py
Skir0s: I think this was kinda mentioned in the other review you made (the one i still havent resolved), this is one area of the code thats a bit bodged up to make stuff work and could potentially be completely reworked in the future (or now)

teamshortcut: Ah sorry, hadn't seen the comments you added above. It's worth considering the best way to handle content that exceeds the character limit; at a minimum, there should be a check that truncates the content if it goes over. In more ideal circumstances, some level of pagination could be included, though I don't know how often quotes go over 2000 characters, so I'm not sure if it is worth it if the code is overly complex. Moving numbers to constants for this kind of stuff is probably a good idea regardless.

Skir0s: I'll leave this one open incase anyone has any alternate ideas, or if we should go ahead with how it is set up right now.

thebeanogamer:
image

Skir0s:
closing this for now, gonna spend some time rewriting doesnt work too well rn

PR #395, opened 2020-04-19
Actually start the loop

thebeanogamer:


picapi:
Seems legit - not had a chance to actually test it out but I don't see it breaking into 1000 pieces at least

thebeanogamer:
@AnonGuy needs to make this

PR #396, opened 2020-04-19
Fix the loop

AnonGuy:
This is the same as #395


picapi:
Duplicate of #395 
haha yes

Issue #429, opened 2020-05-04
Maths Abuse Protection


thebeanogamer:
The LaTeX commands probably needs all of these.

* Rate limiting (user, channel, overall buckets)
* Limit the number of equations in a message
* Lock down which channels it can be used in.

picapi:
The second bullet point here is somewhat addressed in #430; however may warrant further amendment.

picapi:
Additional changes to this effect have been included in #437, with more presumably to come.

-

thebeanogamer:
I think we've done as much as this needs, we can re-open if it becomes a problem again.

PR #430, opened 2020-05-04
Make LaTeX renderer less spammable

picapi:
The LaTeX render should only trigger on the first three expressions in a message if this is merged. This is to try and reduce spamming from the bot if someone was to issue requests for multiple expressions at once. We may wish to consider additional measures to this, but for now this should help a little.


picapi:
> Whilst I think we should try and implement the rest of #429, this is a good start

Agreed - just thought to get a quick patch out to reduce the impact of an immediate flaw ;D

PR #433, opened 2020-05-05
Add :someone command

clubby789:
Recreates Discord's April Fools @someone command. Example:
image
Currently chooses a random user from members, although this could be reworked to pull from online members, or those who have sent a message in the channel within the last n messages.


thebeanogamer:
We don't need this... we just don't.

clubby789:
On a side note, is there a good way to recreate all the CI checks locally? I'd rather avoid having to make 5 commits to fix everything in future.

thebeanogamer:
> 
> 
> On a side note, is there a good way to recreate all the CI checks locally? I'd rather avoid having to make 5 commits to fix everything in future.

Remove the last 2 lines of `tox.ini` then run `flake8`.

PR #434, opened 2020-05-05
Rate limit `:latex`

thebeanogamer:
Helps with #429, needs @AnonGuy to look at if this will work for the inline LaTeX.


Sh3llcod3 (approved):

thebeanogamer:
This doesn't work on inline

H4ckerJ4cker:
Yeah this will only rate limit the command, we need to add a rate limit to the bit of code that does the inline latex. This can be done by importing discord.py cooldowns.

H4ckerJ4cker:
```

cd_mapping = CooldownMapping.from_cooldown(3, 60, BucketType.member)

bucket = self.cd_mapping.get_bucket(message)
            retry_after = bucket.update_rate_limit()
            if retry_after:
                user is rate limited, return and maybe send a error msg
            else:
Do the inline stuff as normal```

PR #436, opened 2020-05-09
Use Caching

### LINKED ISSUE #435: Pipeline caching
https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops

No official python or docker support but we can probably force it in.

thebeanogamer:
Resolves #435

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+.github
+.vscode
+azure-pipelines
+.gitignore
+.security.txt
+CODE_OF_CONDUCT.md
+deployment.yaml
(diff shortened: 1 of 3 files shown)


thebeanogamer:
Bodging in Docker caching added 6 minutes to the build, let's see if there's a better way

thebeanogamer:
Ok best I can tell there isn't a better way, we're just going to have to wait for Azure Pipelines to add proper Docker caching.

lightspeedana:
I see how it is Beano, I'm no longer important to your PR merging :(

PR #446, opened 2020-05-24
Fix inline LaTeX

### LINKED ISSUE #442: Inline LaTeX broken
`$`s are broken. 

This is because of slightly botched blocked channels.

thebeanogamer:
Resolves #442


thebeanogamer:
There's probably a nicer way to do this, but it seems to work

cdbot/cogs/maths.py
```diff
@@ -145,10 +145,16 @@ async def challenge(self, ctx: Context, number: int = 1):
...
+        if type(ctx) is Context:
+            if ctx.channel.id in constants.BLOCKED_CHANNELS:
+                return await ctx.send(
+                    "\N{NO ENTRY SIGN} You cannot use this command in this channel!", delete_after=10
+                )
+        elif type(ctx) is TextChannel:
+            if ctx.id in constants.BLOCKED_CHANNELS:
+                return await ctx.send(
+                    "\N{NO ENTRY SIGN} You cannot use this command in this channel!", delete_after=10
+                )
```
AnonGuy, line 157:
```suggestion
        channel = ctx.channel.id if type(ctx) is Context else ctx.id
        
        if channel in constants.BLOCKED_CHANNELS:
            return await ctx.send(
                "\N{NO ENTRY SIGN} You cannot use this command in this channel!", delete_after=10
            )
```

thebeanogamer: This would be better yes

Issue #428, opened 2020-05-02
Switch LaTeX renderer to quicklatex.com


AnonGuy:

thebeanogamer:
```http
POST /latex3.f HTTP/1.1
Host: www.quicklatex.com

formula=\begin{align*}
x^2 + y^2 %26= 1 \\
y %26= \sqrt{1 - x^2}
\end{align*}&fsize=17px&fcolor=000000&mode=0&out=1&remhost=quicklatex.com&preamble=\usepackage{amsmath}
\usepackage{amsfonts}
\usepackage{amssymb}&rnd=97.07289093482964
```
returns

```text
0
https://quicklatex.com/cache3/82/ql_0676ade0cd04eda37aeb3d0bcd427682_l3.png 0 147 53
```

PR #450, opened 2020-05-27
Switch to https://quicklatex.com/ for LaTeX backend

eohomegrownapps:
# Changes
- We now use https://quicklatex.com/ to render LaTeX.
- All expressions should now render as expected -- x will render as text, whereas $x$ will render as maths.
- The regex to detect inline maths now sends the whole expression to be rendered, rather than rendering each expression enclosed in $$ individually.
- A preamble has been added including packages such as siunitx and mhchem, to allow for rendering of chemical equations and units in physics.
- Compile errors will result in the bot sending an error message, which will disappear after 30 seconds.

# TODOs / nice-to-have features
- The colour looks a little off (this is a carryover from the previous implementation) -- do we want to change this or do something else (e.g. a non-transparent background) to make life easier for both light and dark mode users?
- LaTeX expressions still don't look that great on mobile


AnonGuy:
(closes #450)

thebeanogamer:
Resolves #428

thebeanogamer:
This looks good, would we be able to get some Screenshots please?

cdbot/cogs/maths.py
```diff
@@ -143,31 +142,57 @@ async def challenge(self, ctx: Context, number: int = 1):
...
     @command()
-    async def latex(self, ctx: Context, expression: str):
-        """Render a LaTeX expression."""
+    async def latex(self, ctx: Context, *, expression: str):
+        """
+        Render a LaTeX expression with https://quicklatex.com/
+        """
+        await self.latexRender(ctx, expression)
+
+    async def latexRender(self, ctx: Context, expression: str):
```
AnonGuy, line 151: https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

AnonGuy: Also, why not have the body of `latexRender` in the `latex` method instead?

eohomegrownapps:
Sure :)

math1

math2

thebeanogamer:
I agree about the colour, but other than that it looks good!

cdbot/cogs/maths.py
```diff
@@ -143,31 +142,55 @@ async def challenge(self, ctx: Context, number: int = 1):
...
+
+        preamble = constants.LATEX_PREAMBLE.replace("%","%25").replace("&","%26")
+        
+        body =        'formula=' + formula
+        body = body + '$$$$&fsize=50px'
+        body = body + '&fcolor=969696'
+        body = body + '&mode=0';
+        body = body + '&out=1';
+        body = body + '&errors=1';
+        body = body + '&preamble=' + preamble;
```
AnonGuy, line 175: A couple things worth noting here:

Firstly, `urllib.parse.urlencode` is a handy method for urlencoding parameters. You can use it like this:
```python
from urllib.parse import urlencode

...

body = urlencode({
    'formula': formula + '$$$$',
    'fcolor': 969696,
    'mode': 0,
    'out': 1,
    'errors': 1,
    'preamble': constants.LATEX_PREAMBLE
})
```
Secondly, semicolons aren't required in Python so you can omit those.

eohomegrownapps:
I could just make the colour #FFFFFF which looks good in dark mode... but renders maths invisible in light mode which isn't great :/

cdbot/cogs/maths.py
```diff
@@ -143,31 +142,55 @@ async def challenge(self, ctx: Context, number: int = 1):
...
-            "auth": {"user": "guest", "password": "guest"},
-            "latex": expression,
-            "resolution": 900,
-            "color": "969696",
-        }
+        # Code and regexes taken from https://quicklatex.com/js/quicklatex.js
+        # aiohttp seems to URL-encode things in a way quicklatex doesn't like
+
+        if expression.startswith("...latex") or expression.startswith(":latex"):
+            return
```
AnonGuy, line 163: When invoked with the command, `expression` won't contain the command prefix. You can remove these lines.

cdbot/cogs/maths.py
eohomegrownapps: The issue with this is for some strange reason quicklatex.com doesn't like URL-encoded parameters...

eohomegrownapps: semicolons fixed :) ~was copying from the JS implementation, to work out how they got round the URL-encoding issue...~

cdbot/cogs/maths.py
eohomegrownapps: The issue is that, in order to get the entire text of the user's message rather than have discord split it up into arguments, we need to include the `*` parameter in `latex` -- so it felt easier to have this call a shared rendering function

cdbot/cogs/maths.py
eohomegrownapps: the issue here is that, if those lines are removed, latex_render will fire twice on an expression like `:latex here's some maths $x^2$` -- once because of the `:latex` prefix (the function will be called with `here's some maths $x^2$`) and again because of the inline $ detector regex (the function will be called with `:latex here's some maths $x^2$`). I don't want to remove the :latex function, as it can be handy if the user wants to have something in LaTeX that doesn't include any $ (e.g. align blocks), so this felt like the easiest way to stop the function from firing twice

thebeanogamer:
Have we validated that if someone gets a ping into the error message, it wont execute it?

I'd rather not have a repeat of that time that the bot started spamming @.everyone, even if we have now removed that permission.

thebeanogamer:
Imagine you didn't see me close that

eohomegrownapps:
@thebeanogamer 
> Have we validated that if someone gets a ping into the error message, it wont execute it?
> 
> I'd rather not have a repeat of that time that the bot started spamming @.everyone, even if we have now removed that permission.

I think this is sorted -- the compile error messages appear in an embed (which shouldn't ping anyone), and just to be safe I stripped all @ characters from the compile error messages.

cdbot/cogs/maths.py
Sh3llcod3: So what's happening with this then?

Sh3llcod3:
@thebeanogamer Can you fix the labeller?

thebeanogamer:
> @thebeanogamer Can you fix the labeller?

I keep trying, I think it's a GitHub Actions issue that I need to raise with them. For the moment though, it doesn't need to pass for the PR to be mergable

cdbot/cogs/maths.py
eohomegrownapps: @Sh3llcod3 it feels to me as though the easiest solution is to keep it as-is

Sh3llcod3:
> I keep trying, I think it's a GitHub Actions issue that I need to raise with them. For the moment though, it doesn't need to pass for the PR to be mergable

Alright, there is also a ghost flake8 issue that needs fixing, lint has some problem clearly, but Azure pipelines won't tell you what, cloning locally won't tell you either, that needs to be addressed I think

Sh3llcod3:
@thebeanogamer Help image

flake8_pls

thebeanogamer:
https://dev.azure.com/cyberdiscovery/Cyber%20Discovery%20Discord%20Bot/_build/results?buildId=1430&view=ms.vss-test-web.build-test-results-tab

thebeanogamer:
It won't tell you locally because it's following the tox.ini and outputting a junit file

eohomegrownapps:
oh also do people have thoughts about the colour? would it be better to force the PNG to have a discord-dark-mode-grey background and white (#FFF) text? (rather than as it is currently -- transparent background and greyish white text)

thebeanogamer:
> oh also do people have thoughts about the colour? would it be better to force the PNG to have a discord-dark-mode-grey background and white (#FFF) text? (rather than as it is currently -- transparent background and greyish white text)

I'm not a massive fan of hard coding the background colour like that, but I'm struggling to come up with a better colour combination.

eohomegrownapps:
Background added:

thebeanogamer:
@AnonGuy can you re-review please

PR #453, opened 2020-05-28
Update Deps and fix underlying pep8 issues

Sh3llcod3:
- Update lockfile
- Fix flake8 hopefully


Sh3llcod3:
Never mind, said issues are deeper than I thought

thebeanogamer:
As politely as possible, what in earth are you doing?

Sh3llcod3:
@thebeanogamer If flake8 thinks sentry and discord belong in the same import group and other such non-logical errors, that means there is a serious problem with the image/CI/CI config/version of flake8 itself, it works fine locally. This is a problem :sob: 
flake8_go_brrrrr

Sh3llcod3:
> As politely as possible, what in earth are you doing?
- A) Updating lockfile
- B) Trying to fix the broken flake8/Azure Pipelines issue

thebeanogamer:
A) Dependabot handles that
B) I would seriously dispute that it's broken

Sh3llcod3:
> A) Dependabot handles that
> B) I would seriously dispute that it's broken

For A I understand that, but it doesn't get everything and B is almost definitely broken, can you see why its failing lint? A locally clean installed flake8 with 3.8.3 doesn't find any issues and I can't see anything either. Azure's flake8 is suggesting illogical things like additional newlines which need to be there, combining import groups which are distinctly different, and other things which make no sense. Now if you play along with it, it breaks even harder and starts suggesting the correct things. You make the corrections and bam, it's back to being broken. Locally it works fine.

Sh3llcod3:
You know what, this needs some looking into, but for now I'll close this.

PR #455, opened 2020-05-29
Add Veteran 2018 to Elite Count

thebeanogamer:


cdbot/cogs/cyber.py
```diff
@@ -408,7 +408,11 @@ async def elitecount(self, ctx: Context):
         """
         if ELITECOUNT_ENABLED:
             preferences = {
+                "2018": {
+                    "Attendees": Roles.Elite.VET2018.ATTENDEES
```
AmusedNetwork, line 412: I think to keep it consistent, this line should end with a `,`

thebeanogamer: He's right you know

PR #461, opened 2020-05-31
Automatic readme updating

### LINKED ISSUE #413: Automate Readme Deployment
**Is your feature request related to a problem? Please describe.**
The current #readme deployment commands are fiddly and no one really knows how to use them.

**Describe the solution you'd like**
When the bot boots, it should check if there have been changes to `readme.json` then deploy if there have been. We should probably add some validation though rather than just checking Git, as if the bot reboots and the most recent commit changes #readme, it will redeploy when it doesn't need to.

**Describe alternatives you've considered**
We don't update #readme very often, so we could just leave it as is.

emtee14:
Resolves #413


Sh3llcod3:
Would this work without a persistent disk?

emtee14:
Could try just reading from the readme channel itself and then putting it into a JSON format and then comparing maybe I'm not sure I'll have a think tomorrow

emtee14:
Thinking about it I could get the sha-1 of the file locally then use the GitHub API to get the sha of the file on there then compare it

Sh3llcod3:
Yeah perhaps, my idea was to pin the hash in the dev testing channel and compare the pinned hash with one calculated at deployment.

If same, nothing happens.

If different, then it removes the previous pin, posts and pins the new hash, nukes the channel and creates.

emtee14:
Yh that probs a better idea as I just realised the API one wouldn't work and whats the dev testing channel?

Sh3llcod3:
It's an internal channel for testing bot features, but the point is, any channel like this can be used

emtee14:
oh right cool ill do it that way

emtee14:
This should do the trick the only thing it needs changing is the channel id for the dev testing channel

Sh3llcod3:
Channel ID is: 543766802174443531

emtee14:
Thanks

Sh3llcod3:
Lmk when if this is ready/tested please.

emtee14:
Yh it's ready and tested

Sh3llcod3 (approved):
If the lint issues can be fixed then this should work

emtee14:
I’m not sure why the issues exist because there are plenty of other comments in the file with the same format

emtee14:
Lint issue fixed turns out I was looking at the wrong file should be good to go

Sh3llcod3 (approved):

jerbob (approved):
Not a fan of the camelCase but if it passes linting...

cdbot/cogs/cyber.py
```diff
@@ -494,6 +448,94 @@ async def on_message(self, message: Message):
...
+        hashMsg = None
+        for msg in messages:
+            if msg.content[:6] == "/hash/" and msg.author == self.bot.user:
+                fileHash = msg.content[6:]
+                hashMsg = msg
+                break
+
+        if readmeHash == fileHash:
+            pass
+        else:
```
AnonGuy, line 470: Why not have `if readmeHash != fileHash`?

emtee14: idk ive just fixed it

cdbot/data/readme.json
```diff
@@ -197,4 +197,4 @@
             "color": "0xC1EAC0"
         }
     }
-}
+}
```
thebeanogamer, line 200: Put the trailing newline back please

cdbot/constants.py
```diff
@@ -129,6 +129,10 @@ class Exchange:
 README_SEND_ALIASES = ["create", "push", "generate", "send", "make", "build", "upload"]
 README_RECV_ALIASES = ["fetch", "get", "pull", "download", "retrieve", "dm", "dl"]
 
+# Readme auto update constants
+README_CHANNEL_ID = "409853512185282561"
```
thebeanogamer, line 133: Can we make these overridable with environment variable like the other variables in `constants.py` please

cdbot/cogs/cyber.py
```diff
@@ -494,6 +448,92 @@ async def on_message(self, message: Message):
...
+        readmeFile = "cdbot/data/readme.json"
+        with open(readmeFile, "rb") as f:
+            bytes = f.read()
+            readmeHash = hashlib.sha256(bytes).hexdigest()
+
+        fileHash = None
+        testChannel = await self.bot.fetch_channel(DEV_TESTING_CHANNEL_ID)
+        messages = await testChannel.history().flatten()
+        hashMsg = None
+        for msg in messages:
```
thebeanogamer, line 462: This is a linear search of a potentially massive channel, it's gonna be sssslllloooowwwwww. Perhaps it would make more sense to use the channel description for developer testing, as that will always be a fairly constant length? Similar to what we do for the maths webhook.
image

thebeanogamer (changes requested):

cdbot/cogs/cyber.py
Sh3llcod3: I thought the idea was to use the first pinned message to avoid any searching 🤔

emtee14: I'll do the pin message method then

cdbot/data/readme.json
emtee14: Done

cdbot/cogs/cyber.py
thebeanogamer: > I thought the idea was to use the first pinned message to avoid any searching altogether thinking

What?

cdbot/constants.py
emtee14: Like QUOTES_CHANNEL_ID?

thebeanogamer: Yeah
```suggestion
README_CHANNEL_ID = int(environ.get("README_CHANNEL_ID", "409853512185282561"))
```

cdbot/constants.py
```diff
@@ -129,6 +129,10 @@ class Exchange:
...
+DEV_TESTING_CHANNEL_ID = "543766802174443531"
```
thebeanogamer, line 134:
```suggestion
DEV_TESTING_CHANNEL_ID = int(environ.get("DEV_TESTING_CHANNEL_ID", "543766802174443531"))

```

cdbot/constants.py
emtee14: Ok thanks

cdbot/cogs/cyber.py
emtee14: it would only work if there arent any other pinned messages actually

emtee14: does the dev testing already have a topic or would it just have the hash?

emtee14: Ive got it working where it grabs it from the channel topic

Sh3llcod3: > > I thought the idea was to use the first pinned message to avoid any searching altogether thinking
> 
> What?

We don't have persistent disk, so the idea was pin the hash initially, and for subsequent deploys, compare against the hash in the last pinned message, but channel topic also works I suppose

emtee14: Is there a topic existing that I should add it onto or just leave it as the hash?

Sh3llcod3:
Is this ready now?

emtee14:
Yh it’s just depends whether you want to have a topic for the dev channel aswell as the hash

cdbot/cogs/cyber.py
thebeanogamer: The current topic is empty

emtee14: Ok then this all works fine

emtee14:
Yh this is ready for production I've tested it

emtee14:
Shall i make it auto nuke the channel when the pull command is called?

Sh3llcod3:
Yeah that sounds good

Sh3llcod3:
@thebeanogamer @AnonGuy Can we get a review?

emtee14:
Will this ever be merged?

Sh3llcod3:
No

emtee14:
Ah ok

jchristgit (approved):

PR #463, opened 2020-06-03
Allows Author to Delete Latex

emtee14:


cdbot/cogs/maths.py
```diff
@@ -209,6 +210,16 @@ async def latex_render(self, ctx: Context, expression: str):
                 )
                 return await ctx.send(embed=embed, delete_after=30)
 
+    @Cog.listener()
+    async def on_raw_reaction_add(self, payload):
+        if ord(str(payload.emoji)[0]) == 128465:
```
AnonGuy, line 215:
```suggestion
        if payload.emoji == "\N{WASTEBASKET}":
```

cdbot/cogs/maths.py
```diff
@@ -209,6 +210,16 @@ async def latex_render(self, ctx: Context, expression: str):
...
+            channel = await self.bot.fetch_channel(channel_id=payload.channel_id)
+            msg = await channel.fetch_message(payload.message_id)
+            filename = msg.attachments[0].filename
+            user = await self.bot.fetch_user(payload.user_id)
+            if str(user).replace("#", "") == filename[:-4]:
```
AnonGuy, line 220: Why replace the `#`?

emtee14: because when the file name is saved for some reason it doesn't include the # in the name so user#2843 would be user2843

emtee14: `if user.name + user.discriminator == filename[:4]:` this would work tho aswell

PR #465, opened 2020-06-15
Setup Code Scanning

thebeanogamer:


thebeanogamer:
This adds 3 minutes to the build, but if it's actually helping with security than I'm happy to do so.

PR #470, opened 2020-06-22
Prevent labeler running on forked repositories

### LINKED ISSUE #469: PR Labeler
I like the PR labeler, but because of GitHub Action's permissions, it breaks on PRs from Forks. We should perhaps look at changing the workflow so it only runs on PRs that aren't from forks. See actions/labeler#36

AnonGuy:
Resolves #469


thebeanogamer:
More seriously, this may not work, but it shouldn't break anything if it doesn't.

PR #471, opened 2020-06-22
fix latex deletion

thomas-wigley:


thomas-wigley:
On second thoughts this probably won't fix it.

PR #472, opened 2020-06-22
Fix latex author deletion

dnzc:
Fix my crappy code crappily


thebeanogamer:
Can we get a gif of this working please

dnzc:
Finally - the problem is that when invoked with $$$$ only, the context is type TextChannel rather than Context which is the type when invoked as an actual command ie ...latex or :latex - now the first line of the latex_render function makes sense but there could of at least been a comment explaining it and saving hours of hair tearing... finding a fix now

dnzc:
ezgif-4-439ad8951159

dnzc:
It should work now (yes I did a typo when typing `tests` in the gif)

cdbot/cogs/maths.py
```diff
@@ -144,15 +144,26 @@ async def challenge(self, ctx: Context, number: int = 1):
...
+    async def latex_render(self, ctx: Context, message: Message):
+
+        # When invoked with the word latex, it is run as a command so ctx is type Context
+        # but when invoked with just $$$$ the context is a TextChannel
+        # so we need to check ctx's type
         channel = ctx.channel.id if type(ctx) is Context else ctx.id
 
+        # get the latex expression to be rendered
+        expression = message.content
+        if message.content.startswith("...latex "):
```
thebeanogamer, line 162: I know we did this before but :vomiting_face:, a better solution should be found. @AnonGuy thoughts?

dnzc: Is it really that bad? The message needs to be passed to the function when inline or it can't be stored for checking anyway, so passing in the message as well as the message content seems redundant. And about the if, that can be made a one liner if you want (there's a similar if below so I thought you guys were ok with it). `...latex \LaTeX` still works if you want a gif of that... I might be a noob but is there a particular reason this is bad code?

dnzc: passing the expression to `latex_render` as well as the message does seem like a good way to to it though, it you really don't like this.

cdbot/cogs/maths.py
```diff
@@ -144,15 +144,26 @@ async def challenge(self, ctx: Context, number: int = 1):
...
+            expression = message.content.replace("...latex ", "")
+        elif message.content.startswith(":latex "):
+            expression = message.content.replace(":latex ", "")
```
AnonGuy, line 165:
```suggestion
expression = message.content

if type(ctx) is Context:
    channel = ctx.channel.id
    for prefix in "...latex ", ":latex ":
        expression = expression.replace(prefix, "")
else:
    channel = ctx.id
```

cdbot/cogs/maths.py
```diff
@@ -203,21 +214,24 @@ async def latex_render(self, ctx: Context, expression: str):
...
+                rendered_message = await ctx.send(file=File(image_bytes, filename="result.png"))
+                await rendered_message.add_reaction("\N{WASTEBASKET}")
 
                 # checks if the person who reacted was the original latex author and that they reacted with a bin
                 def should_delete(reaction: Reaction, user: Member):
-                    return ctx.message.author == user and reaction.emoji == "🗑️"
+                    is_right_message = reaction.message.id == rendered_message.id
+                    is_right_user = message.author == user
+                    is_right_emoji = reaction.emoji == "\N{WASTEBASKET}"
+                    return is_right_message and is_right_user and is_right_emoji
```
AnonGuy, line 225:
```suggestion
                    return all((
                        message.author == user,
                        reaction.emoji == "\N{WASTEBASKET}",
                        reaction.message == rendered_message,
                    ))
```

cdbot/cogs/maths.py
thebeanogamer: Remember that the bot can also be invoked by tagging it at the start of a message.

AnonGuy: Would prefer it if there was some context method to strip prefixes but meh

AnonGuy: I am sure there's a cleaner way to handle this (possibly by not calling the same function for commands and non-commands?) but that can go in a different PR I'm sure
image

dnzc:
okay, i removed the find/replace and now its like before but the message itself is also passed in - any better?

thebeanogamer:
🤔 
image

thebeanogamer:
I'm trying to work out why this isn't working, it isn't logging, and it seems to work for you...

dnzc:
arghhh if you add prints do they give an indication of the error?

thebeanogamer:
Interestingly, this seems to still raise the exception, even if I do add the bucket emoji

image

thebeanogamer:
Ok updating my Python venv has fixed the bucket not being removed

thebeanogamer:
The check seems to always return false...

thebeanogamer:
https://github.com/CyberDiscovery/cyberdisc-bot/blob/99c6ed1d16e146e0127bf05d435c7160dba02521/cdbot/cogs/maths.py#L208-L214

If I replace this with `return True` it works, so the problem is with this function

thebeanogamer:
Specifically `reaction.message == rendered_message`, which is weird given that I can see they are equal in the debugger

thebeanogamer:
Ok my debugger is now unsure `rendered_message` actually exists in that context

dnzc:
w
h
y
i
t
n
o
w
o
r
k

cdbot/cogs/maths.py
```diff
@@ -203,21 +202,25 @@ async def latex_render(self, ctx: Context, expression: str):
...
+                rendered_message = await ctx.send(file=File(image_bytes, filename="result.png"))
+                await rendered_message.add_reaction("\N{WASTEBASKET}")
 
                 # checks if the person who reacted was the original latex author and that they reacted with a bin
                 def should_delete(reaction: Reaction, user: Member):
-                    return ctx.message.author == user and reaction.emoji == "🗑️"
+                    return all((
+                        message.author == user,
+                        reaction.emoji == "\N{WASTEBASKET}",
+                        reaction.message == rendered_message,
```
Skir0s, line 213:
```suggestion
                        reaction.message.id == rendered_message.id,
```

PR #473, opened 2020-06-23
Add prow-actions labeler

AnonGuy:
As per [this one comment on the issue](https://github.com/actions/labeler/issues/12#issuecomment-647813044).


thebeanogamer:
:thinking:  This just runs on a schedule rather than PR creation

thebeanogamer:
Hmm, they do use Cron in the official example.

https://github.com/jpmcb/prow-github-actions/blob/20878b551191c551b313a068366e2ec99f220494/.github/workflows/prLabeler.yml#L2-L4

AnonGuy:
yeah, that's why I used the cron snippet in the first place. Not sure if it's cron fixing it

thebeanogamer:
Yeah, I'd stick with their approach

Issue #294, opened 2019-11-25
Integrated Administration Functionality


thebeanogamer:
So I know this has come up before ~and I've been against it~, but I think that the point has come where we have started to grow beyond what Dyno can offer, and a plan has started to be ~feature creeped~ discussed in #staff. We'd be interested to see community feedback, suggestions, and cries of "You'll never actually build it" whilst uncomfortably glancing at the activities website...

Suggestions this far (not all from me, I'm just collating them):
* Discord OAuth
* Role based access control
* Case tracking
* Staff voting on cases and the bot taking moderation access as appropriate
* Web interface for both administration and users wanting to view ongoing moderations
* Allow moderators to attach notes to users
* Bot posts moderation actions as they are taken
* Modmail

Obviously, some of these ideas are better than others, and some are more doable than others, but we'd be interested to see any further ideas or opinions on the ones suggested here!

NicholasG04:
All of them sound like a waste of time to me. Just scrap Dyno, get Gaius Cicereius, don't worry about a shitty web panel it's a waste of time. No voting should be necessary on 'cases' since in experiences I've seen action has to be taken there and then.

thebeanogamer:
The point of this is less replacing automod, it's more to focus on situations where actual humans are required to make choices.

NicholasG04:
I don't see the point of an overly complex webapp for a simple task. What the server really needs imo are punishments for warnings and better guidelines on when to warn people. All of this stuff about cases really seems ott for me

thebeanogamer:
Oh, I agree, it's overkill as hell, but equally, overkill as hell is kinda our thing. Plus, standardising moderation would be assisted by this, as it provides greater visability into cases, as well as presets which can be used by moderators.

Sh3llcod3:
So I think I was writing something for this, but this is definitely more defined as to what it needs...

lightspeedana:
Here's a lovely list of automod features we should be able to add ourselves:

- Delete banned words, with an automute 
- Delete and warn for fast message spam
- Delete and warn for invite links 
- Delete and warn for mass mentions, followed by automute (max count 8)

There's also selfbot detection, but I don't know if that's implementable and will leave that up to you. 

Automute time is 15 minutes, and you need 3 violations of the above to receive one (unless stated otherwise). Link cooldown needs to be 60 seconds. 

I hope this is all implementable and clear 😄 Happy to further clarify if needed.

[...]

PR #475, opened 2020-06-26
[closes #359] Add quote deletion logging

H4ckerJ4cker:
When a quote gets deleted the bot also sends a log message to the logging channel which includes the original quote and a list of the users the quote was deleted by.

PR #476, opened 2020-06-27
Automod for #294

H4ckerJ4cker:
Started adding automod functionality much like dyno. https://github.com/CyberDiscovery/cyberdisc-bot/issues/294#issuecomment-630912015

Things that still need to be added:

- Anti fast messages
- Auto mute after 3 infractions
- Create a banned word list (and decide where it goes)
- Add logging


thebeanogamer:
Please use Git properly, you've not branched from origin/master correctly so this PR also includes all of #475

thebeanogamer:
Also why is this in `fun.py`?

PR #489, opened 2020-07-09
Rework :flag

### LINKED ISSUE #481: Send a cheating video if someone does :flag (or on a regex match)
We've just found a CD cheating video that pops up on a duplicate flag.
Send a link to this in addition to the random string on the `:flag` command, and maybe add it if a message contains the words `flag` and `sell`, or something similar.
A command could also be added to return this (such as `!cheating`)

Video link:
https://game.joincyberdiscovery.com/assets/videos/cheating_message.mp4

clubby789:
Now has a 1 in 5 chance to send the cheating video instead when the :flag command is used.  Basic implementation of #481 . Might make more sense to have the link in a constant but here for now.

Resolves #481


cdbot/cogs/cyber.py
```diff
@@ -167,10 +167,18 @@ async def flag(
             content = "13.1 is a No Flag Zone™ 🙅⛔⚔️"
         else:
             # Generates random, but unique and identical per challenge, base 64 "flag"
-            content = (
-                "The flag is:"
-                f"||{await generatebase64(ord(base[0]) + level_num + challenge_num)}||"
-            )
+            if random.randint(1, 5) == 5:
+                content = (
```
thebeanogamer, line 171: Content is part of an embed. I know you can't put videos in embeds, but just posting the link to it within an embed would look rubbish. Maybe drop the embed entirely if this happens, so Discord expands the link?

clubby789: Just `ctx.send(content)` work okay?

thebeanogamer: That would probably work, although at that point you've made a single use variable, so just do `ctx.send("")`

clubby789: Not sure if `ctx.send()` followed by `return` is best practice but seems like the cleanest way of sending and returning.

cdbot/cogs/cyber.py
```diff
@@ -167,10 +167,20 @@ async def flag(
             content = "13.1 is a No Flag Zone™ 🙅⛔⚔️"
         else:
             # Generates random, but unique and identical per challenge, base 64 "flag"
-            content = (
-                "The flag is:"
-                f"||{await generatebase64(ord(base[0]) + level_num + challenge_num)}||"
-            )
+            if random.randint(1, 5) == 5:
+
+                await ctx.send(
```
Skir0s, line 172: the call to return and await can be combined.
```suggestion
                return await ctx.send(
```

cdbot/cogs/cyber.py
```diff
@@ -167,10 +167,20 @@ async def flag(
...
+                    "The flag is:"
+                    "||https://game.joincyberdiscovery.com/assets/videos/"
+                    "cheating_message.mp4?version=4.2.0||"
+                    # Line break to make PEP8 happy
+                )
+                return
```
Skir0s, line 178:
```suggestion
```

cdbot/cogs/cyber.py
```diff
@@ -167,10 +167,19 @@ async def flag(
...
-                "The flag is:"
-                f"||{await generatebase64(ord(base[0]) + level_num + challenge_num)}||"
-            )
+            if random.randint(1, 5) == 5:
+
+                return await ctx.send(
+                    "The flag is:"
+                    "||https://game.joincyberdiscovery.com/assets/videos/"
+                    "cheating_message.mp4?version=4.2.0||"
+                    # Line break to make PEP8 happy
```
AnonGuy, line 176: No
```suggestion
```

AnonGuy (dismissed):

clubby789:
Is there anything left that needs to be changed?

thebeanogamer (changes requested):
I don't think that just putting the video link in an embed looks very good, whereas if it was outside if it would unfurl. Indeed, if the link is the only thing in the message, the message is hidden and only the video is shown.

clubby789:
Not quite sure what you mean. 
```py

return await ctx.send(
                    "The flag is: "
                    f"||{CHEATING_VIDEO}||"
                )
```
No embed is sent. As for the other part, just change it to send the URL, and nothing else?

thebeanogamer (approved):

PR #491, opened 2020-07-11
Update :elitecount to reflect individual course counts as well

lightspeedana:
I have rejigged some of the formatting slightly as well, so visibility is improved when others want to do the same code. I've added 500, 503, 504 and EHF to the elitecount thing. My only concern is formatting as we'll have a way larger embed than before, but I haven't seen the embed running yet to determine this. Hopefully this is mergeable and works well.

### DIFF
--- a/cdbot/constants.py
+++ b/cdbot/constants.py
@@ -84,7 +84,6 @@ class Roles:
             ELITE500 = int(environ.get("2020_500_MEMBERS_ID", "719957182308679822"))
             ELITEEHF = int(environ.get("2020_EHF_MEMBERS_ID", "722188749378683020"))
 
-
     class Exchange:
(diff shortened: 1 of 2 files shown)


AnonGuy:
[A few linting issues that need addressing](https://dev.azure.com/cyberdiscovery/Cyber%20Discovery%20Discord%20Bot/_build/results?buildId=1616&view=ms.vss-test-web.build-test-results-tab)

cdbot/constants.py
```diff
@@ -96,6 +93,7 @@ class Exchange:
 BOT_TOKEN = getenv("BOT_TOKEN")
 SENTRY_URL = getenv("SENTRY_URL")
 
+
```
thebeanogamer, line 96:
```suggestion
```

lightspeedana:
<3 beano

PR #497, opened 2020-07-19
Update Kubectl

thebeanogamer:


lightspeedana:
How much will this break 😛

PR #505, opened 2020-07-26
Block discord.com invites

AnonGuy:


cdbot/constants.py
```diff
@@ -113,7 +113,7 @@ class Exchange:
 ROOT_ROLE_ID = int(environ.get("ROOT_MEMBERS_ID", "450113490590629888"))
 SUDO_ROLE_ID = int(environ.get("SUDO_MEMBERS_ID", "450113682542952451"))
 ADMIN_ROLES = ("Root", "Sudo")
-BANNED_DOMAINS = ["discord.gg"]
```
thebeanogamer, line 116: I don't think we actually ever call this.

AnonGuy: true

AnonGuy: I changed it, checked for usage and figured there's little point changing it back

thebeanogamer: Yeah, may as well

PR #507, opened 2020-08-01
added user join and leave messages

H4ckerJ4cker:
I don't see why we use dyno for this when it takes about 4 lines.


cdbot/cogs/general.py
```diff
@@ -34,6 +37,17 @@ async def on_ready(self):
             f"\nAuthor: **{latest.author}** on {date}"
         )
 
+    @Cog.listener()
+    async def on_member_join(self, member):
+        join_msg_channel = self.bot.get_channel(WELCOME_CHANNEL)
```
thebeanogamer, line 42: Single use variable

cdbot/cogs/general.py
```diff
@@ -34,6 +37,17 @@ async def on_ready(self):
...
+        join_msg = await join_msg_channel.send(f"{member.mention}, {WELCOME_MESSAGE}")
+        await join_msg.add_reaction('👋')
+
+    @Cog.listener()
+    async def on_member_remove(self, member):
+        leave_msg_channel = self.bot.get_channel(WELCOME_CHANNEL)
```
thebeanogamer, line 48: Another single use variable

cdbot/constants.py
```diff
@@ -100,6 +100,11 @@ class Exchange:
 LOGGING_CHANNEL_ID = int(environ.get("LOGGING_CHANNEL_ID", "538494690601992212"))
 WELCOME_BOT_ID = 155149108183695360
 
+# General constants
+WELCOME_MESSAGE = "Welcome to the Cyber Discovery discussion discord! Please check the rules, roles and information " \
```
thebeanogamer, line 104: This is not how you deal with lines being too long in Python, it should be 
```py
WELCOME_MESSAGE = ("Welcome to the Cyber Discovery discussion discord! Please check the rules, roles and information "
"in <#409853512185282561> to answer any questions, and before you begin.")
```

cdbot/constants.py
```diff
@@ -100,6 +100,11 @@ class Exchange:
...
+                  "in <#409853512185282561> to answer any questions, and before you begin. "
+WELCOME_CHANNEL = 411573884597436416
```
thebeanogamer, line 106: This should be an environment variable to allow for easier debugging

thebeanogamer (changes requested):
Whilst I think this technically will work, you can cast a `member` to `str`, but the code needs a lot of work before merging.

cdbot/cogs/general.py
rak1507v4: This isn't a problem, it's to prevent unnecessarily long lines. Readability counts.

cdbot/cogs/general.py
rak1507v4: See above.

cdbot/cogs/general.py
```diff
@@ -34,6 +37,17 @@ async def on_ready(self):
             f"\nAuthor: **{latest.author}** on {date}"
         )
 
+    @Cog.listener()
+    async def on_member_join(self, member):
+        join_msg_channel = self.bot.get_channel(411573884597436416)
```
thebeanogamer, line 42: This is the opposite of the requested change

thebeanogamer (changes requested):

H4ckerJ4cker:
Because custom bots are nice, I feel like it makes us look better if a user is greeted by our own bot not some generic one.

thebeanogamer (approved):
Well it worked in the screenshots

PR #535, opened 2020-09-14
A command which returns a random fact

csharpisez:
A command which returns a random fact (using the randfacts package)

### DIFF
--- a/cdbot/cogs/fun.py
+++ b/cdbot/cogs/fun.py
@@ -9,6 +9,7 @@ from random import randint
 from string import ascii_lowercase
 from typing import List
 from urllib.parse import urlencode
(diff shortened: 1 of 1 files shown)


thebeanogamer:
What? Why? No.

david-cooke:
why

david-cooke:
@thebeanogamer @Bentechy66 @AnonGuy thoughts on adding this to ractf backend?

Bentechy66:
@david-cooke I think our users would benefit lots from a "fact of the day" box, how educating!

H4ckerJ4cker:
_I can get behind this fully_. This would be beneficial to any project.

PR #556, opened 2020-10-14
Poll system

H4ckerJ4cker:
This does exactly the same thing as what the other bot does, so I coded it for our bot. Its Hacktoberfest baby! for this to work a webhook has to be created and the id set as an environment var. Emojis can also be configured with environment vars.


Sh3llcod3 (commented):
Has this been tested?

H4ckerJ4cker:
That shouldn't be a question @Sh3llcod3 
image

Sh3llcod3 (approved):

KyokoKirigiri11037:
OOPSIE WOOPSIE!! Uwu We make a fucky wucky!! A wittle fucko boingo! The code monkeys at our headquarters are working VEWY HAWD to fix this!

JoshHeng:
Maybe it could be better if the webhook is created & fetched automatically by the bot itself, allowing for one less environment variable and easier deployment?

H4ckerJ4cker:
As in make new one every time and then delete it? @JoshHeng

JoshHeng:
> 
> 
> As in make new one every time and then delete it? @JoshHeng

No, you should be able to fetch all webhooks for a certain channel and send to them. So look in the channel for a `CyberDiscovery Bot` named webhook or something and create it if it doesn't exist.

david-cooke:
doesnt unbelievaboat make this entirely redundant

H4ckerJ4cker:
Customness is cool @david-cooke and its hacktoberfest baby! Everything counts

H4ckerJ4cker:
@JoshHeng Don't think adding that is worth it as it would just be extra code that is run every time the bot starts when it doesn't really have too. It's not that much more effort to make a webhook for developing. The benefit of having one less thing in constants isn't worth it for the mostly abundant code that will be significantly more.

cdbot/constants.py
```diff
@@ -99,6 +99,11 @@ class Exchange:
 QUOTES_BOT_ID = 292953664492929025
 LOGGING_CHANNEL_ID = int(environ.get("LOGGING_CHANNEL_ID", "538494690601992212"))
 WELCOME_BOT_ID = 155149108183695360
+POLL_CHANNEL_ID = int(environ.get("POLL_CHANNEL_ID", "574878887625097227"))
+POLL_WEBHOOK_ID = int(environ.get("POLL_WEBHOOK_ID"))
+POSITIVE_EMOJI = str(environ.get("POSITIVE_EMOJI", "<:this:586246261552840705>"))
+NEUTRAL_EMOJI = str(environ.get("NEUTRAL_EMOJI", "<:somewhatthis:639201446356713502>"))
+NEGATIVE_EMOJI = str(environ.get("NEGATIVE_EMOJI", "<:notthis:485489378999926794>"))
```
thebeanogamer, line 106: This is just worse...

cdbot/constants.py
```diff
@@ -99,6 +99,11 @@ class Exchange:
 QUOTES_BOT_ID = 292953664492929025
 LOGGING_CHANNEL_ID = int(environ.get("LOGGING_CHANNEL_ID", "538494690601992212"))
 WELCOME_BOT_ID = 155149108183695360
+POLL_CHANNEL_ID = int(environ.get("POLL_CHANNEL_ID", "574878887625097227"))
+POLL_WEBHOOK_ID = int(environ.get("POLL_WEBHOOK_ID"))
```
thebeanogamer, line 103: a) Your PR should note that this env var needs to be added
b) Because this is a secret, it should use `getenv` instead

```suggestion
POLL_WEBHOOK_ID = int(getenv("POLL_WEBHOOK_ID"))
```

cdbot/cogs/fun.py
```diff
@@ -577,6 +582,24 @@ async def baldj(self, ctx: Context, *, text: str):
         """
         await self.create_text_image(ctx, "AgentJBadHairDay", text)
 
+    # Polls
+    @command()
+    async def suggest(self, ctx: Context, *, poll_question: str):
```
thebeanogamer, line 587: Should have a docstring

cdbot/cogs/fun.py
```diff
@@ -577,6 +582,24 @@ async def baldj(self, ctx: Context, *, text: str):
...
+        confirm_msg = Embed(
+            description=f":white_check_mark: Your suggestion has been sent to <#{POLL_CHANNEL_ID}> to be voted on.",
+            color=0x6ABE6C)
+        confirm_msg.set_author(name=ctx.message.author.name, icon_url=ctx.message.author.avatar_url)
+        await ctx.send(embed=confirm_msg)
+
+        poll_webhook = await self.bot.fetch_webhook(POLL_WEBHOOK_ID)
```
thebeanogamer, line 594: Unused assignment

thebeanogamer (dismissed):

cdbot/constants.py
H4ckerJ4cker: a is addressed in the the first comment

cdbot/cogs/fun.py
H4ckerJ4cker: I don't think it is...

cdbot/constants.py
H4ckerJ4cker: set to 🟥 and 🟨... by default but can be configured to use any emoji by setting it in the env vars.

cdbot/cogs/fun.py
fractalpenguin: `poll_webhook` is used on line 607, @thebeanogamer

david-cooke:
why is this in my notifications again

omkarmaheshprabhu:
please asign this issue this is reason why iam a good person

PR #585, opened 2020-11-01
Case insensitive help command

H4ckerJ4cker:
Any suggestions welcome. This subclasses the function in discord.py that gets cogs for the help command so it can match case insensitively. In this way :help Fun will work but also :help fun


H4ckerJ4cker:
ah shucks, forgot to do a git pull somewhere, don't think its messed anything up tho

cdbot/cogs/help.py
```diff
@@ -1,11 +1,20 @@
...
+from discord import Colour, Embed, utils
 from discord.ext import commands
 
 
 class EmbeddedHelpCommand(commands.HelpCommand):
     def __init__(self):
         super().__init__(command_attrs={'help': 'Gives detailed information about a command.'})
 
+    async def command_callback(self, ctx, *, command=None):
+        bot = ctx.bot
```
AmusedNetwork, line 10: Since bot is only used once, perhaps all instances of bot should be replaced with `ctx.bot`.

cdbot/cogs/help.py
```diff
@@ -1,11 +1,20 @@
...
+        if command is not None:
+            if (cog := utils.find(lambda c: c.casefold() == command.casefold(), bot.cogs)) is None:
+                cog = command
+            await super().command_callback(ctx, command=cog)
+        else:
+            await super().command_callback(ctx, command=None)
```
AmusedNetwork, line 16: Could both of the `await super()` calls be moved out of the `if command is not None` statement, and then replace `command=cog`, and `command=None` with `command=None`?

cdbot/cogs/help.py
```diff
@@ -1,11 +1,20 @@
...
 from discord.ext import commands
 
 
 class EmbeddedHelpCommand(commands.HelpCommand):
     def __init__(self):
         super().__init__(command_attrs={'help': 'Gives detailed information about a command.'})
 
+    async def command_callback(self, ctx, *, command=None):
+        bot = ctx.bot
+        if command is not None:
```
AmusedNetwork, line 11: Could `if command is not None` be `if command` instead?

AmusedNetwork (changes requested):

H4ckerJ4cker:
ok, this is very bad, but it works and its fast, the same thing can be achieved with a few if statements but we love one liners don't we.

H4ckerJ4cker:
This can be merged safely, worked on a it with people in general and this seems to be the right balance between readability and amount of code.

cdbot/cogs/help.py
H4ckerJ4cker: yes and also no

thebeanogamer: Truthiness go brrrrr

Sh3llcod3 (approved):

H4ckerJ4cker:
No don't close kek, this is good fix

thebeanogamer (approved):
Probably fine

PR #590, opened 2020-11-14
Added *ALL* of the custom commands.

pwnker:
Some commands are hidden from the help command. Let me know if you to change that in one or all commands. Role commands are not part of this as they are done in my rank command pr #587 . As always lmk of any improvements or optimisations.


Sh3llcod3 (approved):
I have no idea if this is needed or not, but sure, I guess. jacc

H4ckerJ4cker:
> I have no idea if this is needed or not, but sure, I guess. jacc

not really needed but it does mean that all the fun and response commands have the same prefix.

Sh3llcod3:
jacc-o-boy

H4ckerJ4cker:
Also I moved like 5 things over and it was merged before I found out there were wayyy more. No point doing half a job.

Sh3llcod3 (approved):

Sh3llcod3 (approved):

thebeanogamer (approved):
Now will you stop asking

Issue #369, opened 2020-04-02
Raid Protection


thebeanogamer:
**Is your feature request related to a problem? Please describe.**
Gaius has repeatedly failed to detect and block raids, we need to redo this.

**Describe the solution you'd like**
Detect high rates of joins? This needs suggestions so it doesn't overreact to things like the flood of users we get when Assess opens.

ZomBMage:
Should be easy enough, autotrigger on a certain number of joins in 30 seconds, then do some funky stuff

thebeanogamer:
Year 2 particularly, we exceeded that for legit users though. I suppose that's a boundary test...

thebeanogamer:
I suppose if it was high enough that could work though.

ZomBMage:
Just make the command enable/disable-able? So if we can preempt lots of users joining at once, like Elite camps, we can just disable it for a time

thebeanogamer:
`:lockdown` could also be an option, kill all invites and the permission to generate invites? Then `:reopen` to generate a new invite and hit the Cloudflare API to update the short URL?

Bottersnike:
I think bulk invite killing would probably be better on a differently named command. `:lockdown` would suggest more that new users are locked out of all channels and have no read/write perms anywhere until the server is unlocked again.

thebeanogamer:
Multiple staff member turnkey perhaps?

thebeanogamer:
> 
> 
> I think bulk invite killing would probably be better on a differently named command. `:lockdown` would suggest more that new users are locked out of all channels and have no read/write perms anywhere until the server is unlocked again.

Yeah, this is a good point

Sh3llcod3:
So with some discussion with Picapi he proposes a system that hits these points:

- Some sort of analysis system where new members can be assigned a score and based on that redirected to some captcha, some system to slow them down and verify that they're a legitimate user

- Some sort of system where in case of mass joins, which is detected as suspicious, the system turns off the Dyno join/leave messages

- Some way for Sudos to revoke invites/re enable them

JamesVStone:
I have seen some bots that generate a captcha and DM it to the user. You could just setup `:captcha on/off` to help prevent raids.

Sh3llcod3:
This sounds interesting 👀

Skiller9090:
Or we could just have a time based system where you do the command:
:userlimit <number>
[...]

PR #610, opened 2020-12-29
Adds basic raid protection.

404dcd:
This "starts the ball rolling" on issue #369 - a simple toggle command that toggles the ability of @.everyone to create instant invites, and on the ON toggle will delete all invites. It also sends a log to the logging channel. It is a root only command.

The code does work, but I'm making it a draft for now because I think we can add some extra functionality. Please voice your suggestions.


cdbot/cogs/admin.py
```diff
@@ -89,6 +91,50 @@ async def on_member_join(self, member: Member):
             # assign placeholder nickname
             await member.edit(nick=PLACEHOLDER_NICKNAME)
 
+    @command()
+    @has_role(ROOT_ROLE_ID)
```
Sh3llcod3, line 95: But there are only two roots as of now 🤔

cdbot/cogs/admin.py
```diff
@@ -89,6 +91,50 @@ async def on_member_join(self, member: Member):
...
+            embed.description = ("Raid protection now ON - All invite "
+                                 "links were deleted and members may not "
+                                 "create new ones")
+        else:
+            embed.description = ("Raid protection now OFF - Members can now "
+                                 "create new invite links")
+        embed.set_author(name=ctx.author.name, icon_url=ctx.author.avatar_url)
+        await logs_channel.send(embed=embed)
+
+        if raidon:
```
Sh3llcod3, line 134: Could this be placed below the first `if raidon` clause below line 127 instead? So it preps the embed and revokes the roles all under one if statement?

Sh3llcod3 (approved):
Good so far, maybe minor changes.

cdbot/cogs/admin.py
404dcd: So allow Sudos as well?

cdbot/cogs/admin.py
404dcd: Yep, I can do that now

H4ckerJ4cker:
Instead of yeeting all the invites we should probs have a system in place where new members are given a default role with perms to view channels and deny it from @everyone. If raid mode is on then don't give members that role. And only show them a channel explaining what raid mode is

cdbot/cogs/admin.py
Sh3llcod3: Idk, you'd probably have to ask staff since that'd be their view, but I'm just saying from an availability standpoint :+1: :joy:

H4ckerJ4cker:
Instead of yeeting all the invites maybe we can think of a system that gives users a role that has perms to view the server on join and when raid mode is on dont give that role

H4ckerJ4cker:
What about the default invite thats mapped to that domain beano has

404dcd:
> What about the default invite thats mapped to that domain beano has

So, that can either be reinstated manually or we could add some functionality that hits Cloudflare's API to update it.

thebeanogamer (commented):
This looks like it'll work, I'll approve it properly once I've tested it

Sh3llcod3 (approved):

thebeanogamer (changes requested):
Ok I've now tested this. I'm happy that it works, but there's a couple small tweaks I'd like before merging.

1. Can we put the admin's name in the audit log please. I know it goes to `#log`, but having it in both would be nice
2. I'm not a massive fan of having this as a toggle. I think it would work better as the base command showing you the current value, then you having to specify off/on. The current model has too high a chance for accidently toggling it to the wrong value because you don't know the current value.

404dcd:
Ok, I'll get right on it.

Sh3llcod3 (approved):
Do it

thebeanogamer (approved):

PR #663, opened 2021-11-23
updated game data for 2021

kastatine:
no volcano 😭


thebeanogamer (approved):
:shrug: I can't confirm if this is accurate as I no longer have access to CyberStart, but this looks reasonable

thebeanogamer:
Sadge

PR #669, opened 2021-12-24
Auto add reactions to quotes

pwnker:


Sh3llcod3 (approved):
I don't see a problem code-wise not sure if @thebeanogamer wants this?

thebeanogamer:
We don't need this. Quotes get a max of about 3 downvotes and the upvotes aren't used for anything. Most people forgot this feature even existed until it was pointed out today.

thebeanogamer:
We should probably reduce the deletion quota back to 5 though, I'm happy to merge that PR

pwnker:
Maybe they would get more downvotes if reactions were auto added as the feature would be used intuitively. With regards to upvotes, they were added here just as a way to "like" a particular quote.

Sh3llcod3:
@pwnker wanted to reopen this yes

